PR #13, opened 2015-05-12
Use `_count` instead of `_total` when exposing summaries

bcandrea:
@grobie: as per the [exposition format spec](https://docs.google.com/document/d/1ZjyKiKxZV83VI9ZKAXRGKaUKK2BIWCT7oiGBKDBpjEY/edit#), summaries exposed in text format should use the suffix `_count` instead of `_total`:

> The type summary is difficult to represent in the text format. The following conventions apply:
>   - Each quantile x is given as a separate sample, each with a label {quantile="_x_"}.
>   - The sample sum for a summary named x is given as a separate sample named _x\_sum_.
>   - The sample count for a summary named x is given as a separate sample named _x\_count_.

### DIFF
--- a/lib/prometheus/client/formats/text.rb
+++ b/lib/prometheus/client/formats/text.rb
@@ -55,7 +55,7 @@ module Prometheus
 
           l = labels(set)
           yield metric("#{name}_sum", l, value.sum)
-          yield metric("#{name}_total", l, value.total)
+          yield metric("#{name}_count", l, value.total)
         end
 
         def self.metric(name, labels, value)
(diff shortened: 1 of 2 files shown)


beorn7:
This looks reasonable, but I'm sure somebody looked at this before. @grobie, what is the context here?

bcandrea:
I might be wrong, but the fact that the value object uses `total` seems to come from the name of its attribute (which I understand is not `count` to avoid overriding the method from `Hash`).

beorn7:
That could be the issue... Let's definitely wait for @grobie to clarify. I think he is still traveling.

grobie:
Thanks a lot @bcandrea. I don't have any explanation for this bug. Your theory sounds plausible.

At SoundCloud almost all ruby projects got moved to JRuby and use the [Java client](https://github.com/prometheus/client_java). That's the explanation why this hasn't been detected earlier.

grobie:
This commit is part of the 0.4.2 release which I should release to rubygems.org: https://rubygems.org/gems/prometheus-client

Changelog: https://github.com/prometheus/client_ruby/compare/v0.4.1...v0.4.2

PR #17, opened 2015-09-11
Remove Counter#decrement

### LINKED ISSUE #16: Counters can be decremented
The prometheus documentation states that:

> Counters should not be used to expose current counts of items whose number can also go down, e.g. the number of currently running goroutines.

Currently counters can be decremented by `decrement` or passing a negative value into `increment`.

grobie:
It was decided that counters should always be monotonically increasing,
so we have to remove possibilities to decrement a counter.

Fixes #16.

@jc @juliusv

### DIFF
--- a/lib/prometheus/client/counter.rb
+++ b/lib/prometheus/client/counter.rb
@@ -11,12 +11,10 @@ module Prometheus
       end
 
       def increment(labels = {}, by = 1)
+        fail ArgumentError, 'increment must be a non-negative number' if by < 0
+
         label_set = label_set_for(labels)
         synchronize { @values[label_set] += by }
-      end
-
-      def decrement(labels = {}, by = 1)
-        increment(labels, -by)
       end
 
       private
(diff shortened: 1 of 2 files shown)


lib/prometheus/client/counter.rb
```diff
@@ -11,14 +11,12 @@ def type
       end
 
       def increment(labels = {}, by = 1)
+        fail ArgumentError, 'increment must be a postive integer' if by < 1
```
brian-brazil: It must be non-negative, 0 and 0.5 are both fine.

grobie: Hm. Right, actually it's about floats (damn untyped languaged). I don't see too much sense in increasing by 0, but fair enough given it's the same as not increasing at all.

grobie: Fixed.

spec/prometheus/client/counter_spec.rb
```diff
@@ -31,6 +31,12 @@
       end.to change { counter.get }.by(5)
     end
 
+    it 'raises an ArgumentError on non-positive increments' do
```
brian-brazil: non-negative

brian-brazil:
:+1: 

There's the odd use case, but it's more for consistency with Summary/Histogram.

Issue #21, opened 2016-05-17
Add Histogram support


grobie:
See https://prometheus.io/docs/concepts/metric_types/#histogram

grobie:
Solved with #24 thanks to @rolandhawk.

PR #24, opened 2016-07-11
histogram support for ruby client

rolandhawk:
Is this acceptable?

ref #21

### DIFF
--- a/lib/prometheus/client/histogram.rb
+++ b/lib/prometheus/client/histogram.rb
@@ -1,0 +1,73 @@
+# encoding: UTF-8
+
+require 'prometheus/client/metric'
+
+module Prometheus
+  module Client
+    # A histogram samples observations (usually things like request durations
+    # or response sizes) and counts them in configurable buckets. It also
+    # provides a sum of all observed values.
+    class Histogram < Metric
+      # Value represents the state of a Histogram at a given point.
+      class Value < Hash
+        attr_accessor :sum, :total
+
+        def initialize(buckets)
+          @sum = 0.0
+          @total = 0
+
+          buckets.each do |bucket|
+            self[bucket] = 0
+          end
+        end
+
+        def observe(value)
+          @sum += value
+          @total += 1
+
+          each_key do |bucket|
+            self[bucket] += 1 if value <= bucket
+          end
+        end
+      end
+
+      # DEFAULT_BUCKETS are the default Histogram buckets. The default buckets
+      # are tailored to broadly measure the response time (in seconds) of a
+      # network service. (From DefBuckets client_golang)
+      DEFAULT_BUCKETS = [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1,
+                         2.5, 5, 10].freeze
+
+      # Offer a way to manually specify buckets
+      def initialize(name, docstring, base_labels = {},
+                     buckets = DEFAULT_BUCKETS)
+        raise ArgumentError, 'Unsorted buckets, typo?' unless sorted? buckets
+
+        @buckets = buckets
+        super(name, docstring, base_labels)
+      end
+
+      def type
+        :histogram
+      end
+
+      def observe(labels, value)
+        if labels[:le]
+          raise ArgumentError, 'Label with name "le" is not permitted'
+        end
+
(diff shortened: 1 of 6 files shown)


spec/prometheus/client/formats/text_spec.rb
```diff
@@ -67,6 +80,13 @@
 qux{for="sake",code="1",quantile="0.99"} 15.3
 qux_sum{for="sake",code="1"} 1243.21
 qux_count{for="sake",code="1"} 93
+# TYPE xuq histogram
+# HELP xuq xuq description
+xuq{code="ah",le="10"} 1
+xuq{code="ah",le="20"} 2
+xuq{code="ah",le="30"} 2
```
brian-brazil: You must expose a +Inf bucket wit the same value as _count

lib/prometheus/client/histogram.rb
```diff
@@ -0,0 +1,59 @@
...
+
+          each do |bucket, count|
+            self[bucket] += 1 if value <= bucket
+          end
+        end
+      end
+
+      # Offer a way to manually specify buckets
+      def initialize(name, docstring, base_labels = {}, buckets = [])
+        @buckets = buckets
```
brian-brazil: What are the default buckets?

lib/prometheus/client/histogram.rb
```diff
@@ -0,0 +1,59 @@
...
+          @total += 1
+
+          each do |bucket, count|
+            self[bucket] += 1 if value <= bucket
+          end
+        end
+      end
+
+      # Offer a way to manually specify buckets
+      def initialize(name, docstring, base_labels = {}, buckets = [])
```
brian-brazil: You need to check the buckets are ordered.

rolandhawk: Do I need to validate buckets order and raise if not ordered 
or
Just order the buckets?

brian-brazil: We've raised elsewhere, as it's likely a typo on the part of the user.

lib/prometheus/client/histogram.rb
```diff
@@ -0,0 +1,64 @@
...
+      def initialize(name, docstring, base_labels = {}, buckets = DEFAULT_BUCKETS)
+        @buckets = buckets.sort
+        super(name, docstring, base_labels)
+      end
+
+      def type
+        :histogram
+      end
+
+      def add(labels, value)
```
brian-brazil: This should be called `observe`

lib/prometheus/client/registry.rb
```diff
@@ -43,6 +44,10 @@ def gauge(name, docstring, base_labels = {})
         register(Gauge.new(name, docstring, base_labels))
       end
 
+      def histogram(*args)
```
grobie: For documentation purposes, I'd vote to explicitly list the arguments.

lib/prometheus/client/histogram.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      # DEFAULT_BUCKETS are the default Histogram buckets. The default buckets are
+      # tailored to broadly measure the response time (in seconds) of a
+      # network service. (From DefBuckets client_golang)
+      DEFAULT_BUCKETS = [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10]
+
+      # Offer a way to manually specify buckets
+      def initialize(name, docstring, base_labels = {}, buckets = DEFAULT_BUCKETS)
+        @buckets = buckets.sort
```
grobie: I agree with @brian-brazil and would vote to validate and raise an error on unsorted buckets. It's easy to make a typo when specifying small buckets values. We should help the user to detect such issues.

grobie:
Thanks for your contribution! This looks great in general, please only fix the last few comments.

I updated the formatting to the latest rubocop rules (not sure following rubocop was a good idea, they change their mind so often ...). Please rebase and squash your changes against current master. I'll fix the test dependencies for "old" ruby versions tomorrow.

rolandhawk:
Hi @grobie @brian-brazil, I have updated my PR, please kindly check it

grobie:
Thanks a lot @rolandhawk!

PR #28, opened 2016-08-15
use histogram for http_request_duration_seconds

### LINKED ISSUE #27: Use histogram in rack collector.
Seeing histogram support land is great. Can the Rack collector be updated to export it?

tmc:
closes #27


grobie:
So the tests would need to get updated, but I was thinking about changing the default metric name according to @brian-brazil's http://www.robustperception.io/on-the-naming-of-things/. Given this will be a breaking change anyway, we could do both things at once.

How awkward is the namespacing? Would it make more sense to move the rack middlewares to something like `Prometheus::Rack::Exporter`?

grobie:
Replaced by #32

PR #31, opened 2016-09-08
Add support for DELETE request

mikeurbach:
@grobie please review

it would be handy for my apps if the DELETE api was exposed by the ruby client


lib/prometheus/client/push.rb
```diff
@@ -36,6 +36,10 @@ def replace(registry)
         request('PUT', registry)
       end
 
+      def delete
+        request('DELETE')
```
grobie: Given we're not going to send the registry, I guess we shouldn't pretend that in the Content-Type header. Wdyt about just calling `@http.send_request(...)` directly here?

mikeurbach: makes sense to me, i will update it

grobie:
Thanks, LGTM. Please add a small stub to the readme example as well.

mikeurbach:
@grobie i just realized i accidentally checked in a version of the gem i had built for testing... do you want me to make a PR or will you remove it?

grobie:
I saw that and understood it as an accident, but it doesn't really hurt
either. I'll bump it correctly once I release the new version, don't worry
about it.

mikeurbach:
oh, ya i was uploading this to a private gem server for testing. but i mean i actually checked in the built gem to the codebase: https://github.com/prometheus/client_ruby/blob/master/prometheus-client-0.7.0.beta.gem

grobie:
Oooh, shoot. Thanks for the heads-up, I removed it.

Issue #10, opened 2015-02-09
Missing protobuf support


grobie:
There is currently no support for prometheus' protobuf format. The format is described here: http://prometheus.io/docs/instrumenting/exposition_formats/. 

The protobuf definition itself has been created already: https://github.com/prometheus/client_model/tree/master/ruby.

beorn7:
Just as a side note: The new Java client ("simple client") doesn't support protobuf yet, either. To vet the urgency of adding the support, we ran some benchmarks to find out how much network and server resources the different formats require. The result was that protobuf parsing is much faster than the text format (which is in turn a bit faster than JSON parsing). Uncompressed, the protobuf format is smaller than the text format, but compressed, there is not much difference. However, if you apply compression, the relative difference between protobuf and text format it terms of parsing effort shrinks as the de-compression step takes a similar amount of time for both formats. While the gain of the text format over the deprecated JSON format seems small in comparison, it is enough to get down to a resource usage where other things become more important. Even with pure text format clients, you'll probably be able to support most of the use cases. Only for very high throughput cases (tens of thousands of samples ingested per sec), it will matter if you use protobuf or not.
So it is still an improvement to support protobuf, but other things might be more urgent to implement first.

grobie:
Protobuf support has been removed in Prometheus 2.0

PR #32, opened 2016-09-09
Improve rack middlewares

### LINKED ISSUE #27: Use histogram in rack collector.

grobie:
- Move from prometheus/client/rack/\* to prometheus/middleware/*
- Use histogram instead of summary in collector middleware
- Spepicfy collector metrics by changing the prefix to http_server_*
- Remove HTTP Host header from default label builder

Fixes #27.

@tmc and @mikeurbach as users of the library.
@brian-brazil for comments on the default metrics names and help strings.

```
# TYPE http_server_requests_total counter
# HELP http_server_requests_total The total number of HTTP requests handled by the Rack application.
http_server_requests_total{method="get",path="/",code="200"} 74
http_server_requests_total{method="post",path="/",code="200"} 64
http_server_requests_total{method="delete",path="/",code="404"} 18
http_server_requests_total{method="delete",path="/",code="200"} 61
http_server_requests_total{method="get",path="/",code="404"} 8
http_server_requests_total{method="post",path="/",code="404"} 15
# TYPE http_server_request_latency_seconds histogram
# HELP http_server_request_latency_seconds The HTTP response latency of the Rack application.
http_server_request_latency_seconds{method="get",path="/",code="200",le="0.005"} 74
http_server_request_latency_seconds{method="get",path="/",code="200",le="0.01"} 74
http_server_request_latency_seconds{method="get",path="/",code="200",le="0.025"} 74
...
# TYPE http_server_exceptions_total counter
# HELP http_server_exceptions_total The total number of exceptions raised by the Rack application.
http_server_exceptions_total{exception="NoMethodError"} 10
```

I was thinking about removing `http_server_requests_total` as it's included in the histogram, but I've seen beginners to be heavily confused using `http_server_request_latency_seconds_count` as counter. I guess it's not too harmful.


lib/prometheus/middleware/collector.rb
```diff
@@ -0,0 +1,79 @@
...
+        {
+          method: env['REQUEST_METHOD'].downcase,
+          path:   env['PATH_INFO'].to_s,
+        }
+      end
+
+      def init_request_metrics
+        @requests = @registry.counter(
+          :http_server_requests_total,
+          'The total number of HTTP requests handled by the Rack application.',
```
brian-brazil: If it's from rack, then the metric name should contain "rack"

lib/prometheus/middleware/exporter.rb
```diff
@@ -0,0 +1,89 @@
+# encoding: UTF-8
+
+require 'prometheus/client'
+require 'prometheus/client/formats/text'
+
+module Prometheus
+  module Middleware
+    # Exporter is a Rack middleware that provides a sample implementation of
+    # a Prometheus HTTP client API.
```
brian-brazil: I'm pretty sure this isn't implementing a Prometheus HTTP client API. It's exposition.

lib/prometheus/middleware/collector.rb
```diff
@@ -0,0 +1,79 @@
...
+        yield.tap do |response|
+          duration = (Time.now - start).to_f
+          record(labels(env, response), duration)
+        end
+      rescue => exception
+        @exceptions.increment(exception: exception.class.name)
+        raise
+      end
+
+      def labels(env, response)
```
brian-brazil: It's hard to follow the label logic here.

lib/prometheus/middleware/collector.rb
```diff
@@ -0,0 +1,79 @@
...
+        @label_builder.call(env).tap do |labels|
+          labels[:code] = response.first.to_s
+        end
+      end
+
+      def record(labels, duration)
+        @requests.increment(labels)
```
brian-brazil: A http request does not have a response code.

lib/prometheus/middleware/collector.rb
```diff
@@ -0,0 +1,79 @@
...
+        @durations.observe(labels, duration)
```
brian-brazil: Latency should not be broken out by success/failure,  put it in a separate metric.

lib/prometheus/middleware/collector.rb
```diff
@@ -0,0 +1,79 @@
...
+        @exceptions = @registry.counter(
+          :http_server_exceptions_total,
+          'The total number of exceptions raised by the Rack application.',
+        )
+      end
+
+      def trace(env)
+        start = Time.now
+        yield.tap do |response|
+          duration = (Time.now - start).to_f
```
brian-brazil: Can this be negative?

olleolleolle:
Thanks a lot for improving the Rack middleware.

As a novice Prometheus user providing data from a Rack app, it was nice to see a middleware.

As I got it up and presenting data, I later learned that the "summary" metric was wrong for what we'd tried to use it for. We now find ourselves in [the wrong column of table "Histogram vs Summary"](https://prometheus.io/docs/practices/histograms/#quantiles). 

Can we move this forward in any way? Are all of the understanding-improving/renaming comments blocking a merge?

grobie:
That, but mostly it's me traveling for the last couple of months. I'll have
some time between Christmas and new year's and it's on my list to get this
rolling again.

perlun:
@grobie - any updates on this one?

grobie:
I've been traveling in the Amazon jungle for the last few weeks, but will have more internet connections during the next couple of days. I'll try to pick this up again.

lib/prometheus/middleware/collector.rb
grobie: Rack is the de facto standard interface for ruby HTTP servers. It can be implemented by many different servers (unicorn, thin, puma, ...). I thought about including it, but concluded after some discussions that including it in the metric name is neither expected nor helpful.

lib/prometheus/middleware/exporter.rb
grobie: Hm, I see what you mean. Better now?

lib/prometheus/middleware/collector.rb
grobie: Hm, I guess it makes sense to not include it by default as it's easy to misjudge such metrics. I disagree that histograms should never be broken out by status code, as discussed in https://github.com/prometheus/client_golang/pull/250 though and left the option to change the labels.

lib/prometheus/middleware/collector.rb
grobie: Yes, during system clock changes. Unfortunately, there is no standard way to access a monotonic clock across all supported ruby versions, so I opted to just clamp the duration a 0 seconds. Wdyt?

grobie:
The metrics generated by the included HTTP tracer would look like this:

```
# TYPE http_server_requests_total counter
# HELP http_server_requests_total The total number of HTTP requests handled by the Rack application.
http_server_requests_total{code="200",method="get",path="/"} 76.0
http_server_requests_total{code="200",method="post",path="/"} 67.0
http_server_requests_total{code="404",method="delete",path="/"} 16.0
http_server_requests_total{code="200",method="delete",path="/"} 62.0
http_server_requests_total{code="404",method="get",path="/"} 7.0
http_server_requests_total{code="404",method="post",path="/"} 10.0
# TYPE http_server_request_latency_seconds histogram
# HELP http_server_request_latency_seconds The HTTP response latency of the Rack application.
http_server_request_latency_seconds{method="get",path="/",le="0.005"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.01"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.025"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.05"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.1"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.25"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.5"} 83
http_server_request_latency_seconds{method="get",path="/",le="1"} 83
http_server_request_latency_seconds{method="get",path="/",le="2.5"} 83
...
http_server_request_latency_seconds{method="delete",path="/",le="5"} 78
http_server_request_latency_seconds{method="delete",path="/",le="10"} 78
http_server_request_latency_seconds{method="delete",path="/",le="+Inf"} 78
http_server_request_latency_seconds_sum{method="delete",path="/"} 0.003424651000000001
http_server_request_latency_seconds_count{method="delete",path="/"} 78
# TYPE http_server_exceptions_total counter
# HELP http_server_exceptions_total The total number of exceptions raised by the Rack application.
http_server_exceptions_total{exception="NoMethodError"} 12.0
```

brian-brazil:
> The term request is pretty overloaded and I always understood it to describe here the full cycle of a HTTP request and response.

Request means request, response means response.

> It's a very common need to have a metric broken out by request method, path and response code. Neither http_request_... nor http_response_... metrics would include all three. I'm open to suggestions, but the current approach seems to be the best option IMHO.

Neither requests nor latency should be broken out by response code. Requests because that's to do with responses not requests. And latency because those are our guidelines around reporting success/failure with latency.

I'd suggest switching the request counter to being called response, and dropping the code label from the latency.

grobie:
> Request means request, response means response.

And still it's commonly understood that a statement "The request latency was 400ms" includes both the time to sent a HTTP request and receive a HTTP response without the need to say something like "The request and response latency was ...". 

> I'd suggest switching the request counter to being called response, and dropping the code label from the latency.

As commented earlier, I had already dropped the code label from the latency histogram.

Following your argumentation, a `http_server_responses_total` metric wouldn't be allowed to include `method` and `path` labels as a HTTP response doesn't have such fields by definition.

I understand the naming isn't perfect and remember the same discussion with Matt years ago. Considering the colloquial use of the term `request`, the similar problems with labels on a `http_responses_total` metric, the benefit of having the same prefix on all HTTP metrics and the existing usage of `http_requests_total` metrics in various client libraries, I'm for keeping the current naming pattern.

Are there other committers voting to change the naming? @beorn7 @juliusv

beorn7:
The naming used by https://godoc.org/github.com/prometheus/client_golang/prometheus#InstrumentHandler in client_golang should not serve as a role model in any way. It's strongly deprecated (and coming up with a replacement that doesn't immediately run into similar or other problems turns out to be hard).

The problem with the metric name `http_requests_total`, as I understood it so far, is the one described in https://prometheus.io/docs/instrumenting/writing_exporters/#naming : 

> Generally metric names should allow someone who’s familiar with Prometheus but not a particular system to make a good guess as to what a metric means. A metric named http_requests_total is not extremely useful - are these being measured as they come in, in some filter or when they get to the user’s code?

Thus, the name should include some kind of subsystem. (In this case, perhaps `rack_http_requests_total`? But I'm not too familiar with the Ruby code here.)

The discussion around `request` vs. `response` is new to me. I certainly want to see how many of my `DELETE` _requests_ got a `200` _response_.  If I separated those two from each other, how would I do this?

juliusv:
While I've heard about the `request` vs. `response` argument before, I agree that it's often useful to have both the method (or path) and the status code in one metric, and that the colloquial use of "request" is not only about the request, but also the response. @brian-brazil or would there be a better metric name wording?

grobie:
Regarding the inclusion of `rack` in metric names.

> The problem with the metric name http_requests_total, as I understood it so far, is the one described in https://prometheus.io/docs/instrumenting/writing_exporters/#naming :

That's the reason this PR changes the metric names from `http_requests_...` to `http_server_requests_...`. I considered including `rack` in the name as well, but after some thinking and discussions with others opted against it. Copy&pasting my answer to Brian from above:

> Rack is the de facto standard interface for ruby HTTP servers. It can be implemented by many different servers (unicorn, thin, puma, ...). I thought about including it, but concluded after some discussions that including it in the metric name is neither expected nor helpful.

Rack is really just an interface definition and not an implementation in its own. Literally every ruby server uses rack these days, so that servers (unicorn, thin, ...), frameworks (sinatra, rails, ...) and plugins (rack middlewares) can be freely used together. It does not provide any additional information to include `rack` in the metric name, as it's still unclear where the prometheus middlewares are added in the chain of middlewares. For reference, a typical Rails application uses 20-30 middlewares.

beorn7:
The reasoning about the name appears sound to me. However, it also proves how difficult generic naming is. My current plan for the HTTP instrumentation helpers in client_golang is to always let the user provide a name, without default. But as said, we are still in exploration state there.

matthiasr:
> Request means request, response means response.

They are not independent. The response means nothing without the request, and the request has no result without the response.

In an ideal nanoservice world, where each binary does exactly one thing, the responses may make sense without further context, but out here in the real world we _need_ the response breakdown by (method,path,status). We actively graph and alert this, as otherwise high-volume low-value endpoints swamp out signal from low-volume high-value ones. Consider a CRUD service – if the R request rate is 1000x the CUD rate (not uncommon in consumer websites *cough*), and a 0.1% error rate on reads is acceptable, it is impossible to detect a completely failing C/U/D endpoint without method and path.

grobie:
In order to unblock this PR, I'd ask for a vote how to proceed here. Arguments against and for `http_server_requests_total`/`http_server_responses_total` have been written above.

Summary of the changes in this PR:
* changed prefix of all HTTP tracer metrics from `http_...` to `http_server_...`
* replace summary with histogram for `http_server_request_latency` metric
* removed code label from histogram (formerly a summary)
* provided two separate options to change the labels added to `http_server_requests_total` and `http_server_request_latency` (formerly both label sets were changed simultaneously)
* renamed the middlewares from `Prometheus::Client::Rack::XXX` to `Prometheus::Middleware::XXX` to make the breaking changes clear

The newly generated metrics look like this:


```
# TYPE http_server_requests_total counter
# HELP http_server_requests_total The total number of HTTP requests handled by the Rack application.
http_server_requests_total{code="200",method="get",path="/"} 76.0
http_server_requests_total{code="200",method="post",path="/"} 67.0
http_server_requests_total{code="404",method="delete",path="/"} 16.0
http_server_requests_total{code="200",method="delete",path="/"} 62.0
http_server_requests_total{code="404",method="get",path="/"} 7.0
http_server_requests_total{code="404",method="post",path="/"} 10.0
# TYPE http_server_request_latency_seconds histogram
# HELP http_server_request_latency_seconds The HTTP response latency of the Rack application.
http_server_request_latency_seconds{method="get",path="/",le="0.005"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.01"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.025"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.05"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.1"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.25"} 83
http_server_request_latency_seconds{method="get",path="/",le="0.5"} 83
http_server_request_latency_seconds{method="get",path="/",le="1"} 83
http_server_request_latency_seconds{method="get",path="/",le="2.5"} 83
...
http_server_request_latency_seconds{method="delete",path="/",le="5"} 78
http_server_request_latency_seconds{method="delete",path="/",le="10"} 78
http_server_request_latency_seconds{method="delete",path="/",le="+Inf"} 78
http_server_request_latency_seconds_sum{method="delete",path="/"} 0.003424651000000001
http_server_request_latency_seconds_count{method="delete",path="/"} 78
# TYPE http_server_exceptions_total counter
# HELP http_server_exceptions_total The total number of exceptions raised by the Rack application.
http_server_exceptions_total{exception="NoMethodError"} 12.0
```

Please vote with :+1: / :-1: if you agree with these changes. @brian-brazil @beorn7 @juliusv @matthiasr

matthiasr:
:+1:

beorn7:
Slightly different note: We often use `…_duration_seconds` instead of `…_latency_seconds`. I would prefer if one was used consistently (or there are rules when to use what). Does anybody have thoughts about this?

grobie:
I'm totally in favor of using the more common one. Will change!

beorn7:
But _is_ `…_duration_seconds` the more common? I always thought so, but I'm not sure anymore…

brian-brazil:
I don't think there's a clear convention there, I'd say follow whatever the Go client has.

beorn7:
As of now, the Go client has `…_duration_seconds` (actually ``…_duration_microseconds`, but changing that to seconds is one of the easy and obvious changes that will happen).

grobie:
Ok, I changed latency to duration.

beorn7:
:+1: from a pure "how the metrics look like" perspective.

lib/prometheus/middleware/exporter.rb
```diff
@@ -0,0 +1,93 @@
...
+      def initialize(app, options = {})
+        @app = app
+        @registry = options[:registry] || Client.registry
+        @path = options[:path] || '/metrics'
+        @acceptable = build_dictionary(FORMATS, FALLBACK)
+      end
+
+      def call(env)
+        if env['PATH_INFO'] == @path
+          format = negotiate(env['HTTP_ACCEPT'], @acceptable)
```
Strech: I think it will be more efficient to fetch a key from `env`, like `env.fetch('HTTP_ACCEPT', '')` and assume that in `negotiate` method we will always get a string in `accept` argument.

Then we can improve `negotiate` method and replace `accept.to_s.empty?` on `accept.empty?`

Strech: I don't like the idea to set default value here, but it can also be a strategy – always provide `accept` argument, then it could be `env.fetch('HTTP_ACCEPT', '*/*')` or something like `env.fetch('HTTP_ACCEPT', DEFAULT_HTTP_ACCEPT)`, but I'm afraid then it will shift responsibilities of negotiation a bit here

lib/prometheus/middleware/exporter.rb
```diff
@@ -0,0 +1,93 @@
...
+
+        parse(accept).each do |content_type, _|
+          return formats[content_type] if formats.key?(content_type)
+        end
+
+        nil
+      end
+
+      def parse(header)
+        header.to_s.split(/\s*,\s*/).map do |type|
```
Strech: This method is used only by `negotiate` and we will never pass `header` as non-string value, so we can ommit `to_s` here

lib/prometheus/middleware/exporter.rb
grobie: Thanks @Strech! I incorporated your suggestions and simplified the code. Note, most of that was necessary when the ruby client library supported more than one exposition format (json and text). It might become necessary again in the future if #10 gets implemented.

Strech: Copy, I hope that I will have a chance to contribute more in next few months 👍

PR #40, opened 2017-02-08
Issue in example run script

cheynewallace:
Run script throws a `/run: line 38: syntax error near unexpected token >` on OSX


grobie:
This isn't a typo, but a bash feature. See "Appending Standard Output and Standard Error" in the [bash reference](https://www.gnu.org/software/bash/manual/html_node/Redirections.html).

grobie:
Given we redirect to /dev/null anyway, I'm fine with increasing compatibility with old bash versions.

Issue #36, opened 2016-11-03
Implement collector interface


grobie:
The client library should follow our standard collector design and not scrape metrics directly: https://prometheus.io/docs/instrumenting/writing_clientlibs/#overall-structure

grobie:
Once this has been done, more performance optimizations are possible, like incrementing all affected histogram buckets during a scrape instead of during each observe. See https://github.com/prometheus/client_golang/blob/a5060f1eaab721946b185b2de468ff83ea5b89cb/prometheus/histogram.go#L240-L282

dmagliola:
@grobie Reading that doc, I'm not sure I understand what this would be.
How is this different from the `Registry` the client already supports?

As for the performance improvement mentioned, we're doing that already:
https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/client/histogram.rb#L77
https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/client/histogram.rb#L93

:)

nieltg:
@dmagliola, 
I think I can answer from the Go language client library perspective.

`Registry` is able to [register collectors.](https://github.com/prometheus/client_golang/blob/333f01cef0d61f9ef05ada3d94e00e69c8d5cdda/prometheus/registry.go#L167) Collector itself has [two methods:](https://github.com/prometheus/client_golang/blob/333f01cef0d61f9ef05ada3d94e00e69c8d5cdda/prometheus/collector.go#L27-L63) `Describe` and `Collect`. While the scrapping process is running, the handler of `/metrics` endpoint will [gather metrics from the registry](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/http.go#L114-L126) and the registry will [collect all of the registered metrics.](https://github.com/prometheus/client_golang/blob/333f01cef0d61f9ef05ada3d94e00e69c8d5cdda/prometheus/registry.go#L443)

One use case which I encountered here is to create a custom collector that collects metrics when it's instructed. One example on Go lang. is in this LXD exporter which [exports metrics as instructed to.](https://github.com/nieltg/lxd_exporter/blob/master/pkg/metrics/collector.go#L83-L100)

dmagliola:
Thank you for your explanation @nieltg !
I'd like to try to tackle this issue sooner than later, since it unblocks a bunch of other use cases, but i'd like to make sure I understand it fully before starting, to not end up with an inadequate solution.

[...]

Issue #35, opened 2016-11-03
Add Child notion to client


brian-brazil:
Currently labels are handled by being the first argument to all functions. This gives the incorrect impression that all metrics should labels (in reality most metrics don't have labels) and makes label-less use harder.

This client should follow the structure laid out in https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels

In addition the user should be required to specify all their label names at metric creation time.

grobie:
Label-less use does not get harder by having an optional labels parameter. See the first example from the readme. This is a common ruby pattern.

```ruby
http_requests = prometheus.counter(:http_requests, 'A counter of HTTP requests made')

# start using the counter
http_requests.increment
```

Using a labels() method and calling metric functions on that will help performance-wise though. I added this to milestone v0.7.0.

brian-brazil:
> Label-less use does not get harder by having an optional labels parameter. 

The labels parameter is first, so if you want a non-default value it makes it harder.

dmagliola:
@brian-brazil this is now covered as part of the changes we made in PR #95. We require labels at creation time, they're no longer the first parameter, and the Child notion is implemented through the `with_labels` methods.

I'm going to close this issue due to this but feel free to reopen if you think there's still work to do on this.

PR #47, opened 2017-03-03
Remove extra Proc allocation from metrics sync.

Strech:
Tiny improvement on objects allocations


grobie:
Whooops. Thanks for catching that, I just checked with benchmark-memory and it's clearly visible. Also that we do a lot of other unnecessary allocations sadly.

Please note that the library will be completely rewritten to account for #35 (which will help a long way with the performance issues) as well as #36 and following the [client library guidelines](https://prometheus.io/docs/instrumenting/writing_clientlibs) in general.

PR #54, opened 2017-05-05
Validate metric name format

### LINKED ISSUE #37: Metric name validation doesn't match docs
Docs say that the metric name should match `[a-zA-Z_:][a-zA-Z0-9_:]*`: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

The Ruby library just checks that it is a symbol: https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/client/metric.rb#L49

See also https://github.com/prometheus/client_golang/pull/255.

grobie:
Fixes #37

### DIFF
--- a/lib/prometheus/client/metric.rb
+++ b/lib/prometheus/client/metric.rb
@@ -46,9 +46,13 @@ module Prometheus
       end
 
       def validate_name(name)
-        return true if name.is_a?(Symbol)
-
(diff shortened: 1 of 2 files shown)


grobie:
@beorn7

beorn7:
:+1:

PR #57, opened 2017-05-15
Use path normalization in collect label builder example

grobie:
@RichiH So I can just point users to that example in the future.

### DIFF
--- a/examples/rack/README.md
+++ b/examples/rack/README.md
@@ -52,9 +52,10 @@ use Prometheus::Middleware::Collector, counter_label_builder: ->(env, code) {
   {
     code:         code,
     method:       env['REQUEST_METHOD'].downcase,
+    # Include the HTTP Host header as label.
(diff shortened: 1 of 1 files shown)


examples/rack/README.md
```diff
@@ -52,9 +52,10 @@ use Prometheus::Middleware::Collector, counter_label_builder: ->(env, code) {
   {
     code:         code,
     method:       env['REQUEST_METHOD'].downcase,
+    # Include the HTTP Host header as label.
     host:         env['HTTP_HOST'].to_s,
-    path:         env['PATH_INFO'].to_s,
-    http_version: env['HTTP_VERSION'].to_s,
```
RichiH: Did you drop that on purpose?

grobie: Yeah, I thought one additional label in comparison to the default label builder is enough to show what is possible. Happy to add that back if you think it's useful?

PR #62, opened 2017-05-30
Normalize paths including numeric IDs by default

grobie:
Quite a few people run into the issue of creating too many time series
by exporting an unbounded number of paths label values. While it's
documented and possible to provide a custom label builder, a better
default is probably to always normalize ids in HTTP paths.

@SuperQ @brian-brazil


brian-brazil:
:+1:

SuperQ:
Very nice!

PR #65, opened 2017-07-05
Ability to pass custom headers for HTTP requests

av-ast:
I added ability to pass custom headers for HTTP requests.

### DIFF
--- a/spec/prometheus/client/push_spec.rb
+++ b/spec/prometheus/client/push_spec.rb
@@ -21,6 +21,22 @@ describe Prometheus::Client::Push do
       expect(push.gateway).to eql('http://pu.sh:1234')
     end
 
+    it 'allows to specify a custom Pushgateway with custom headers' do
+      headers = { 'Key' => 'Value' }
+      push = Prometheus::Client::Push.new('test-job', nil, nil, headers)
+
+      merged_headers = described_class::DEFAULT_HEADERS.merge(headers)
+      expect(push.headers).to eql(merged_headers)
+    end
+
+    it 'allows to averride default headers' do
+      headers = { 'Content-Type' => 'application/json' }.freeze
+      push = Prometheus::Client::Push.new('test-job', nil, nil, headers)
+
+      expect(push.headers).to eql(headers)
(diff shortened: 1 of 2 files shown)


brian-brazil:
Why do you want this? The pushgateway doesn't use such headers.

av-ast:
In my case I want to pass headers for Basic Auth.

spec/prometheus/client/push_spec.rb
```diff
@@ -21,6 +21,22 @@
...
 
+    it 'allows to specify a custom Pushgateway with custom headers' do
+      headers = { 'Key' => 'Value' }
+      push = Prometheus::Client::Push.new('test-job', nil, nil, headers)
+
+      merged_headers = described_class::DEFAULT_HEADERS.merge(headers)
+      expect(push.headers).to eql(merged_headers)
+    end
+
+    it 'allows to averride default headers' do
```
grobie: What is the use case for setting a different content-type?

av-ast: Actually, there is no real reason to use a different content-type.
I just showed such ability with this test case...

grobie:
I think this change is inline with our principle that clients should support standard authorization when talking to other servers, like our support for scrapes and in federation. Our standard answer to requests for https/etc. for pushgateway and other components is to use a reverse proxy in front of it, therefore clients should have the ability to set required headers. Any concerns with that @brian-brazil?

I'm still not really sure why we should allow to change the Content-Type, but on the other hand, I don't see a strong reason to forbid it.

brian-brazil:
I'd suggest having explicit basic auth support, rather than requiring the user to hand-assemble the Authentication header.

In Python there's a handler you can use for this: https://github.com/prometheus/client_python/#handlers-for-authentication
In Java it's not released yet, but user/password in the passed URL object work. There may be further extensions to support SSL and bearer tokens.

> I'm still not really sure why we should allow to change the Content-Type, but on the other hand, I don't see a strong reason to forbid it.

It's allow the user to set an incorrect content type, the library is choosing the content so it alone should specify the type.

grobie:
Resolved by #74.

PR #66, opened 2017-07-26
Add thread safe increment and decrement operations to gauges

Envek:
Currently, gauges are hard to use as bi-directional counters due to a lack of increment and decrement operations on them.

So I've found myself writing following scary fragments of the code:

```ruby
# Before
simultaneous_service_calls_gauge.send(:synchronize) do
  value = simultaneous_service_calls_gauge.get(tags) || 0
  simultaneous_service_calls_gauge.set(tags, value + 1)
end

simultaneous_service_calls_gauge.send(:synchronize) do
  value = simultaneous_service_calls_gauge.get(tags) || 0
  simultaneous_service_calls_gauge.set(tags, value - 1)
end
```

That's why this PR exists:

```rb
# After
simultaneous_service_calls_gauge.increment(tags)
simultaneous_service_calls_gauge.increment(tags)
```

For now, I've made `increment` and `decrement` to initialize values with `0` before changing (default value of gauge is still `nil`). I've failed to understand the code of python and go client libraries to figure out, whether their gauges have default values or not.

@grobie, can you please take a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ gauge.get({ room: 'kitchen' })
 # => 21.534
 ```
 
+Also you can use gauge as the bi-directional counter:
+
+```ruby
+gauge = Prometheus::Client::Gauge.new(:concurrent_requests_total, '...')
+
+gauge.increment({ service: 'foo' })
+# => 1.0
+
+gauge.decrement({ service: 'foo' })
+# => 0.0
+```
+
 ### Histogram
 
 A histogram samples observations (usually things like request durations or
--- a/lib/prometheus/client/gauge.rb
+++ b/lib/prometheus/client/gauge.rb
@@ -19,6 +19,26 @@ module Prometheus
 
         @values[label_set_for(labels)] = value.to_f
       end
+
+      # Increments Gauge value by 1 or adds the given value to the Gauge.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -129,6 +129,18 @@ gauge.get({ room: 'kitchen' })
 # => 21.534
 ```
 
+Also you can use gauge as the bi-directional counter:
+
+```ruby
+gauge = Prometheus::Client::Gauge.new(:concurrent_requests_total, '...')
+
+gauge.increment({ service: 'foo' })
```
brian-brazil: Per the [spec](https://prometheus.io/docs/instrumenting/writing_clientlibs/#gauge) these should be inc and dev, but this also needs redoign to have the right structure.

Envek: Hmm, I give it the same names as in `Counter` (it have method called `increment` too, which apparently doesn't conform to spec, but it's written 4 years ago, much before the spec, I think)

grobie: Yes, the ruby client codebase is the oldest one and needs quite some rework. I think using increment/decrement for now makes the most sense to keep the codebase itself consistent. It'll all be updated with the rewrite.

grobie:
Thank you!

PR #68, opened 2017-08-29
Support Collector middleware using custom metric names

mpalmer:
It can sometimes be handy to expose the HTTP server metrics under a
different prefix, either for naming consistency (so all your app metrics
have a common prefix), or to prevent conflicts (it is possible, and
occasionally useful, to run more than one Rack-based webserver in a single
process, although it can be a bit of a wrestle).

Includes tests, which pass on at least 2.3 (things seem spectacularly broken on at least 1.9 and 2.0, for unrelated reasons) and doesn't introduce any new rubocop sadnesses.

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -25,7 +25,7 @@ module Prometheus
       def initialize(app, options = {})
         @app = app
         @registry = options[:registry] || Client.registry
-        @metrics_prefix = options[:metrics_prefix] || "http_server"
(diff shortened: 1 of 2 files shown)


mpalmer:
Fixed one cop that slipped through.  I have NFI why coveralls is complaining, the file listing at https://coveralls.io/builds/13044829/source?filename=lib%2Fprometheus%2Fmiddleware%2Fcollector.rb makes no sense to me, and the coverage reports in the test suites all say "Coverage is at 100.0%.".  So I'm working on the assumption coveralls is just unhinged.

grobie:
Thanks!

PR #69, opened 2017-10-09
Normalize UUIDs in request path by default

klippx:
Hello!

We have a lot of microservices written in ruby, and in our ecosystem (which handles tons of transactions every day) we are using UUID's for our entities. Unfortunately, the Ruby client for Prometheus does not by default normalize UUIDs which required us to get closer to the source code of this gem.

One could argue that this should not be core default behaviour of the gem and that as users we should use the option parameter for `counter_label_builder` and `duration_label_builder` instead. We did that, but it is quite a bother going into each microservice and spreading these local overrides. That's why we put it in a forked version of the client, but ideally we would like it to be part of the library itself.

One last thing to consider, do you believe that UUIDs are growing popular in more and more code bases? I think that is the case, so this would be a positive contribution to the library.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,7 +13,3 @@ Style/TrailingCommaInLiteral:
 
 Metrics/AbcSize:
     Max: 18
-
-Metrics/BlockLength:
-    Exclude:
-        - 'spec/**/*.rb'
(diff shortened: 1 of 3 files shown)


klippx:
Comment on failing builds: 
- Looks like Ruby 1.9.3 is failing due to environment/bundler issue. I think I need help kicking Travis to make it green...?
- The coverage issue is probably caused by the failing build since coverage is 100% locally and for other builds

grobie:
Thanks!

PR #70, opened 2017-10-25
Fix threading issue with label_set_for

lethjakman:
This causes an error when parallelized if they're not wrapped together.
Error message: "can't add a new key into a hash during iteration"

### DIFF
--- a/lib/prometheus/client/counter.rb
+++ b/lib/prometheus/client/counter.rb
@@ -13,8 +13,10 @@ module Prometheus
       def increment(labels = {}, by = 1)
         raise ArgumentError, 'increment must be a non-negative number' if by < 0
 
-        label_set = label_set_for(labels)
-        synchronize { @values[label_set] += by }
+        synchronize do
+          label_set = label_set_for(labels)
+          @values[label_set] += by
+        end
       end
 
       private
--- a/lib/prometheus/client/gauge.rb
+++ b/lib/prometheus/client/gauge.rb
@@ -23,8 +23,8 @@ module Prometheus
       # Increments Gauge value by 1 or adds the given value to the Gauge.
       # (The value can be negative, resulting in a decrease of the Gauge.)
       def increment(labels = {}, by = 1)
-        label_set = label_set_for(labels)
         synchronize do
+          label_set = label_set_for(labels)
           @values[label_set] ||= 0
           @values[label_set] += by
         end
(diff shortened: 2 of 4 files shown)


grobie:
@lethjakman When have you seen that behavior? Can you provide a stack trace? The `#label_set_for` function doesn't try to add any key to a hash.

dmagliola:
Hi @lethjakman!
Thank you for your PR! Sorry for the very late response.
This is a valid issue, and one we probably haven't experienced just because we're on MRI.

@grobie `label_set_for` doesn't add to a hash directly, however, it calls the validator, and the validator caches the validation, by adding the labelset to an internal Hash.
As I understand it, that's where this is a problem.
It shouldn't be a problem in MRI (as evidenced by the `concurrent-ruby` gem [using the regular Ruby Hash as a thread-safe one](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/hash.rb#L21), but I can see how it would be in other implementations.

We're recently made a major set of changes to the Client, and we changed the way the Validator works, and I thought we'd gotten rid of that cache, but on a second-look, we haven't.
This PR would need major adaptation to work with the new codebase, however, since the synchronization is now the responsibility of the Data Store, not the Metric.
This would add quite a bit of complexity, to get around an uncommon race-condition on a piece of functionality we don't really need that much.

At this point, I propose we get rid of the [validation cache](https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/client/label_set_validator.rb#L37)
This was always my plan with this code, and I actually thought I had done it already.

The trade-off is obviously performance, although if the users of the gem are impacted by the overhead of validating the same labels over and over, they can now use the `with_labels` method to specify all the labels, which will return a new Metric object that skips the validation (since it can do it once on `initialize`).

We should run this through the performance benchmark and see how it performs, though, to make sure we're not ruining everything, but my memory of working around this code is that it shouldn't be a huge issue.

@Sinjo What do you reckon?

Sinjo:
I'm 👍 on removing the optimisation if it doesn't cause significant<sup>1</sup> performance degradation.

[1] We'll know it when we see it 😛

dmagliola:
Just to update on this... We will fix this problem in a new PR, as this code changed quite a bit. I'll link it here once that's done. Until then, i'll leave this open

dmagliola:
This PR is now replaced by the solution in PR #145, so i'm closing this one, but we can consider this problem solved, once we merge that.

PR #71, opened 2017-11-06
Add summary for request duration quantiles to the collector

klippx:
When setting up monitoring and dashboards for our services we found that we want a convenient way to extract 90 and 99 percentiles for request duration using prometheus primitives without using too much advanced querying on histograms or other indexes.

We think this primitive could be good addition of the basic collector, wdyt?


grobie:
The Prometheus summaries can't be aggregated statistically (like showing the 90th percentile latency for a whole service). That's why histograms exist, see https://prometheus.io/docs/practices/histograms/#quantiles.

Given there will be almost always more than one instance running, I don't think we should export summaries in this collector.

klippx:
Thank you @grobie, that makes perfect sense. 👍

PR #72, opened 2017-11-17
Added synchronize to net/http calls

vsh91:
Net::HTTP is not threadsafe. Unfortanutely, sometimes it causes IOError's during
working on multhreading app. Mutex should solve that problem.


vsh91:
Hello @grobie, could you please review my PR?

PR #74, opened 2017-12-13
client/push: add basic auth

josephholsten:
I should mention that these changes to request make me want to cry. So any alternatives would be much appreciated!


josephholsten:
ok, so I just need to build an assertion that the http double received the right kind of net/http request. ugh.

9to6:
Hello.
Why this PR it not merged yet?
any problems?

grobie:
Fixed in47a6366c45727b2709ac059c6742c31f5369c595.

josephholsten:
Woo, thanks!

PR #81, opened 2018-02-15
Improve gauge documentation example

grobie:
A counter in Prometheus can not go down and should not have the _total
suffix.

@brian-brazil That slipped through somehow.


brian-brazil:
:+1:

PR #82, opened 2018-02-15
Remove RuboCop

grobie:
Github warns about a low security vulernability in the rubocop gem
(which is only used during development). I tried briefly to upgrade the
gem, but I was greeted with hundreds of new style checks. The vast
majority for which RuboCop demanded the opposite in previous versions.

I'm just tired of spending my time to follow the guidelines of whoever
is in charge of RuboCop this year. As much as I'd like to enforce a
consistent code style, the rubocop project doesn't seem to have a clear
idea on such style either.

@brian-brazil @porras


brian-brazil:
:+1: 

I'm not a Ruby person, but that sounds reasonable to me.

porras:
Good call, there might be some value on Rubocop pickiness but most of the time not worth the maintenance effort.

juliancheal:
Instead of removing Rubocop completely, one option is just to run Rubocop on any lines changed (using git diff, etc).

This is something we do over on the ManageIQ project (https://github.com/ManageIQ)

grobie:
@juliancheal I wanted to use Rubocop to ensure consistent code formatting, and fail to see how only enforcing new rules on changed lines will help with that?

dlbock:
@grobie Probably a good idea to upgrade Rubocop to the latest version?

Then you can generate a todo file that you inherit from in `.rubocop.yml` that would contain all the violations that haven't been "fixed" yet, which you can fix at a later time, but doesn't fail the build right now.

I created a gist for the change I made based off `master` and ran `bundle exec rake` and everything passes (specs and rubocop): https://gist.github.com/dlbock/4391150077b4e7c2ed66e7afedff39e4

If you'd rather do that I can submit a PR.

Issue #67, opened 2017-08-10
NoMethodError: undefined method `any?' for #ActionDispatch::Response::RackBody


brahamshakti:
Hi this is my `config.ru` file

    require 'rack'
    require 'prometheus/middleware/collector'
    require 'prometheus/middleware/exporter'

    use Rack::Deflater, if: ->(_, _, _, body) { body.any? && body[0].length > 512 }
    use Prometheus::Middleware::Collector
    use Prometheus::Middleware::Exporter

    run Rails.application

But when I am throwing `:not_found` error using `return head :not_found unless user` it gives me

    #<NoMethodError: undefined method `any?' for #ActionDispatch::Response::RackBody:0x00000003c657c0>>
    /home/user/.rvm/gems/ruby-2.3.3/gems/rack-2.0.3/lib/rack/body_proxy.rb:41:in `method_missing'
    /home/user/.rvm/gems/ruby-2.3.3/gems/rack-2.0.3/lib/rack/body_proxy.rb:41:in `method_missing'
    /home/user/.rvm/gems/ruby-2.3.3/gems/rack-2.0.3/lib/rack/body_proxy.rb:41:in `method_missing'
    /home/user/.rvm/gems/ruby-2.3.3/gems/rack-2.0.3/lib/rack/body_proxy.rb:41:in `method_missing'
    /home/user/api/config.ru:5:in `block (2 levels) in <main>'
    /home/user/.rvm/gems/ruby-2.3.3/gems/rack-2.0.3/lib/rack/deflater.rb:114:in `should_deflate?' 
    /home/user/.rvm/gems/ruby-2.3.3/gems/rack-2.0.3/lib/rack/deflater.rb:37:in `call'
    /home/user/.rvm/gems/ruby-2.3.3/gems/puma-3.8.2/lib/puma/configuration.rb:224:in `call'
    /home/user/.rvm/gems/ruby-2.3.3/gems/puma-3.8.2/lib/puma/server.rb:600:in `handle_request'
    /home/user/.rvm/gems/ruby-2.3.3/gems/puma-3.8.2/lib/puma/server.rb:435:in `process_client' 
    /home/user/.rvm/gems/ruby-2.3.3/gems/puma-3.8.2/lib/puma/server.rb:299:in `block in run'
    /home/user/.rvm/gems/ruby-2.3.3/gems/puma-3.8.2/lib/puma/thread_pool.rb:120:in `block in spawn_thread'`

grobie:
[...]

PR #83, opened 2018-03-01
Updated Rack configuration

jamtur01:
In newer rack versions the if guard fails with a missing method for
`.any?`. The quickest fix is to remove the guard.


jamtur01:
Fixes #67

PR #91, opened 2018-07-23
allow exception label builder to be provided in collector middleware

itszootime:
The ability to customise the labels for the exception counter is currently missing, this PR fixes that 🙂.

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -29,6 +29,7 @@ module Prometheus
         @metrics_prefix = options[:metrics_prefix] || 'http_server'
         @counter_lb = options[:counter_label_builder] || COUNTER_LB
         @duration_lb = options[:duration_label_builder] || DURATION_LB
+        @exception_lb = options[:exception_label_builder] || EXCEPTION_LB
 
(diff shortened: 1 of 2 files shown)


dmagliola:
Hi @itszootime !
Thanks for your PR, and sorry for the late response!
We tried going down this path at GoCardless, and we ended up in a slippery slope full of lambdas, that made the Collector really hard to read.
This became a mess quite quickly, and still, this sort of customization always ended up falling short, and we always ended up re-implementing the collector in our app for one reason or another.
Due to this, we decided to move in the opposite direction instead: We made the collector less configurable, but *a lot* easier to read (at least in my opinion).

With this change, anyone that needs a different implementation can much more easily implement their own (by basically copy/pasting this one, which is what we always ended up doing), or subclassing it (at your own risk, since it's internal API may change). More info [here](https://github.com/prometheus/client_ruby/blob/master/examples/rack/README.md#collector)

I encourage you to take a look at the version of the collector currently in `master` (which is a release candidate version right now, and will soon be the `1.0`), and let us know what you think. 
In the meantime, if that's ok with you, i'll close this PR, since it doesn't match the current version of the code.

PR #92, opened 2018-07-31
Add config.ru example configuration for Rails app

seanfcarroll:


brian-brazil:
Thanks for your PR. As a matter of policy we do not support server-side authentication, and we also consider environment variables unsuitable for secrets - see https://prometheus.io/docs/operating/security/ This example also seems a bit specific to your use case.

What I would suggest is sharing this in a blog post or similar.

seanfcarroll:
Thanks for the feedback, but this is focused towards a Rails app which is hosting the endpoint. If the endpoint is not secured then the data is readable by any user. As for env variables we use they are still pretty common in Rails apps.

Sorry if I'm being daft, but am I missing something here?

Issue #94, opened 2018-10-22
[RFC] Align Ruby client with Prometheus best practices, prepare for multi-process support


dmagliola:
As it currently stands, the Prometheus Ruby Client has a few issues that make it hard to adopt in mainstream Ruby projects, particularly in Web applications:
1. [Pre-fork servers can't report metrics](https://github.com/prometheus/client_ruby/issues/9), because each process has their own set of data, and what gets reported to Prometheus depends on which process responds to the scrape request.
2. The current Client, being one of the first clients created, doesn't follow several of the [Best Practices and Guidelines](https://prometheus.io/docs/instrumenting/writing_clientlibs/). 

We'd like to contribute to the effort of improving the client in these directions, and we're proposing we could make a number of changes (this issue will be promptly followed by a PR that implements several of these).

### Objectives
- Follow [client conventions and best practices](https://prometheus.io/docs/instrumenting/writing_clientlibs/)
- Add the notion of Pluggable backends. Client should be configurable with different backends: thread-safe (default), thread-unsafe (lock-free for performance on single-threaded cases), multiprocess, etc.
  - Consumers should be able to build and plug their own backends based on their use cases.

We have reached out to @grobie recently and he suggested that releasing a new major version was the way to go in order to work around these issues.

There are several proposals in this RFC for improvements to the existing [Prometheus Client for Ruby](https://github.com/prometheus/client_ruby).

These proposals are largely independent of each other, so we can pick for each one whether we think it’s an improvement or not. They are also ordered from most to least important. Only the first one is an absolute must, since it paves the way for adding multi-process support.

# 1. Centralizing and Abstracting storage of Metric values

In the current client, each Metric object has an internal `@values` hash to store the metric values. The value of this hash is, for Gauges and Counters, a `float`, and the key of this hash is itself a hash of labels and their values. Thus, for one given metric there are multiple values at any given time, one for each combination of values of their labels.

For Histograms, `@values` doesn’t hold a `float`. Instead it holds a `Histogram::Value` object, which holds one `integer` for each bucket, plus the total number of observations, and the sum of all observations. Summaries do a similar thing.

We're proposing that, instead of each metric holding their own counter internally, we should have a centralized store that holds all the information. Metric objects update this store for every observation, and it gets read in its entirety by a formatter when the metrics are scraped.

## Why

[...]

Issue #9, opened 2015-02-08
Support pre-fork servers


atombender:
If you use this gem with a multi-process Rack server such as Unicorn, surely each worker will be returning just a percentage of the correct results (eg., number of requests served, total time), thus making the exposed metrics fairly meaningless?

To solve this the easiest solution is to create a block of shared memory in the master process that all workers share, instead of using instance variables.

brian-brazil:
I believe this problem came up with some users of the (experimental) Python client, and some proposed work on the HAProxy exporter.

What you really want to do is separately scrape each worker, and then aggregate it up in the Prometheus server. This gives the most useful/correct results, and also gives you per-process statistics for things like memory and cpu. I'm not familiar with Unicorn or Rack, do they offer anything that'd allow for this?

Shared memory is an option, it has disadvantages though in that it'd be a source of contention and you can't do per-process analysis.

juliusv:
From the [Unicorn Design doc](http://unicorn.bogomips.org/DESIGN.html):

"Unicorn is a traditional UNIX prefork web server. [...] Load balancing between worker processes is done by the OS kernel. All workers share a common set of listener sockets and does non-blocking accept() on them. The kernel will decide which worker process to give a socket to and workers will sleep if there is nothing to accept()."

Given that, I'm not sure how doable it is to scrape individual workers. Even if it was possible, the benefit of having individual insight and avoiding contention has to be weighed against potentially having an order of magnitude more time series.

I'd expect that on average, the workers on a single machine will behave similarly, except admittedly in pathological situations like this one pointed out in the Unicorn docs: "When your application goes awry, a BOFH can just "kill -9" the runaway worker process without worrying about tearing all clients down, just one".

It also states that "Unicorn deployments are typically only 2-4 processes per-core", which would mean multiplying the number of exposed time series on a typical large machine with e.g. 32 cores by 64x to 128x.

@grobie Did you have any plans wrt. supporting preforked servers already?

atombender:
@brian-brazil: It's possible to collect per-process information with shared memory: Just organize the memory by PID. You'll have to reap dead PIDs, though. If you're clever you can avoid locking altogether, by having each process write to its own dedicated segment of the shared memory.

Scraping individual workers, as @juliusv points out, would mostly be useful to the extent that it would capture misbehaving workers that consume too much memory or CPU.

[...]

PR #95, opened 2018-10-22
Abstract Data Storage away from Metric objects, introduce Swappable Data Stores, and support Multiprocess Pre-fork Servers

dmagliola:
This PR attempts to address the first set of changes required for the objectives outlined in [Issue 94](https://github.com/prometheus/client_ruby/issues/94)

This is an excerpt from that issue, please check [the full text](https://github.com/prometheus/client_ruby/issues/94) for more context:

As it currently stands, the Prometheus Ruby Client has a few issues that make it hard to adopt in mainstream Ruby projects, particularly in Web applications:
1. [Pre-fork servers can't report metrics](https://github.com/prometheus/client_ruby/issues/9), because each process has their own set of data, and what gets reported to Prometheus depends on which process responds to the scrape request.
2. The current Client, being one of the first clients created, doesn't follow several of the [Best Practices and Guidelines](https://prometheus.io/docs/instrumenting/writing_clientlibs/). 

### Objectives
- Follow [client conventions and best practices](https://prometheus.io/docs/instrumenting/writing_clientlibs/)
- Add the notion of Pluggable backends. Client should be configurable with different backends: thread-safe (default), thread-unsafe (lock-free for performance on single-threaded cases), multiprocess, etc.
  - Consumers should be able to build and plug their own backends based on their use cases.

------------------

Points this PR tackles:

- A few refactorings / improvements here and there
- Add keyword arguments to methods, to be more idiomatic within modern Ruby standards
- Create the concept of Data Stores, separate from the Metrics the user declares, to have a clear interface that allows us to swap different implementations.
- Introduce 3 data stores, tailored to 3 specific scenarios: Single Threaded, Multi-threaded, and Multi-Process applications.

------------

This PR will be very hard to read if you're looking at all the changes at once. We recommend reading it commit-by-commit, since each one is an incremental step towards the final state, and they have very extensive explanations on the why and how of each step.

I'm sorry about that, it was the only practical way of making this large refactoring. I would've liked for it to be many small, individual PRs, but most of these changes depended on the previous ones.

------------

Short explanation on our rationale for the built-in Multi-Process Data Store being based on Files:
- We experimented with multiple different possible data stores, and we focused strongly on benchmarking them to see which would ones present acceptable performance.
- Since there seems to be a community direction towards using MMaps, we put a good bit of effort on having an `MmapStore`
- We took @juliusv 's efforts in this repo, in the `multiprocess` branch, as a starting point, and adapted it to the Data Stores interface we're presenting here, which was quite easy. They play along well together.
- We then worked on fixing a few stability bugs, and on a few performance improvements with it, and we were quite happy with the results.
- HOWEVER, there is one stability issue we haven't yet been fully able to solve. Under some conditions, this store crashes. We can't reproduce this in our dev machines, but Travis actually crashes frequently.
- We also experimenting taking that exact same approach, but removing the mmap. Basically, it uses files, but indexing them in the same way @juliusv is indexing the mmap, reading and writing the binary-packed Floats directly into their offsets in those files, which is a great idea. This approach is *surprisingly* fast (mostly because of FS caching, we're not really touching the disk for the most part). So, for the time being, we're proposing the `DirectFileStore`, as the official way of working with pre-fork servers.
- Some performance numbers. These is the time to increment a counters, without labels, on a single thread:
  - SingleThreadedStore (basically, simplest store possible, just a hash): 1μs
  - SynchronizedStore (a hash with a mutex around): 4μs
  - MMapStore: 6μs
  - DirectFileStore: 9μs

- So, MMaps are 30% faster, and we consider that enough improvement to continue trying to get that to work, but at 9μs per observation, the DirectFileStore is extremely stable and reliable, and it's pretty fast. 
- Our rationale is that it's better to release this as is, knowing full well that is safe, and that it solves the pre-fork problem for the vast majority of users, rather than rely on a less stable approach, for a performance improvement that'll be important for some users, but not for the majority.

That said, we are preparing a separate repo (https://github.com/gocardless/prometheus-client-ruby-data-stores-experiments) where we're going to dump the rest of the stores we created, more benchmarks, and extensive documentation on all the exploration we did. We think this will be a great starting point for anyone making their ow stores, and we encourage the community to try and help us finalize the MmapStore (or make their own, if they have a better approach), but we don't think all of that belongs in this repo, it'd add a lot of clutter and confusion for consumers of the gem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ before_install:
   - |
     if [[ "$(ruby -e 'puts RUBY_VERSION')" != 1.* ]]; then gem update --system; fi
 rvm:
-  - 1.9.3
   - 2.3.8
   - 2.4.5
   - 2.5.3
--- a/README.md
+++ b/README.md
@@ -214,7 +214,47 @@ https_requests_total = Counter.new(:http_requests_total,
 https_requests_total.increment(labels: { status_code: response.status_code })
 ```
 
+### `with_labels`
 
+Similar to pre-setting labels, you can get a new instance of an existing metric object,
+with a subset (or full set) of labels set, so that you can increment / observe the metric
+without having to specify the labels for every call.
+
+Moreover, if all the labels the metric can take have been pre-set, validation of the labels
+is done on the call to `with_labels`, and then skipped for each observation, which can 
+lead to performance improvements. If you are incrementing a counter in a fast loop, you
+definitely want to be doing this.
+
+
+Examples:
+
+**Pre-setting labels for ease of use:**
+
+```ruby
+# in the file where you define your metrics:
+records_processed_total = registry.counter.new(:records_processed_total, 
+                                               docstring: '...', 
+                                               labels: [:service, :component],
+                                               preset_labels: { service: "my_service" })
+
+# in one-off calls, you'd specify the missing labels (component in this case)
+records_processed_total.increment(labels: { component: 'a_component' })
+
+# you can also have a "view" on this metric for a specific component where this label is
+# pre-set:
+class MyComponent
+  def metric
+    @metric ||= records_processed_total.with_labels(component: "my_component")
+  end
+  
+  def process
+    records.each do |record|
+      # process the record
+      metric.increment 
+    end
+  end
+end
+```
 
 
 ## Data Stores
--- a/README.md
+++ b/README.md
@@ -158,15 +158,18 @@ histogram.get(labels: { service: 'users' })
 Summary, similar to histograms, is an accumulator for samples. It captures
 Numeric data and provides an efficient percentile calculation mechanism.
 
+For now, only `sum` and `total` (count of observations) are supported, no actual quantiles.
+
 ```ruby
 summary = Prometheus::Client::Summary.new(:service_latency_seconds, docstring: '...', labels: [:service])
 
 # record a value
 summary.observe(Benchmark.realtime { service.call() }, labels: { service: 'database' })
 
-# retrieve the current quantile values
-summary.get(labels: { service: 'database' })
-# => { 0.5 => 0.1233122, 0.9 => 3.4323, 0.99 => 5.3428231 }
+# retrieve the current sum and total values
+summary_value = summary.get(labels: { service: 'database' })
+summary_value.sum # => 123.45
+summary_value.count # => 100
 ```
 
 ## Labels
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ The following metric types are currently supported.
 Counter is a metric that exposes merely a sum or tally of things.
 
 ```ruby
-counter = Prometheus::Client::Counter.new(:service_requests_total, docstring: '...')
+counter = Prometheus::Client::Counter.new(:service_requests_total, docstring: '...', labels: [:service])
 
 # increment the counter for a given label set
 counter.increment(labels: { service: 'foo' })
@@ -118,7 +118,7 @@ Gauge is a metric that exposes merely an instantaneous value or some snapshot
 thereof.
 
 ```ruby
-gauge = Prometheus::Client::Gauge.new(:room_temperature_celsius, docstring: '...')
+gauge = Prometheus::Client::Gauge.new(:room_temperature_celsius, docstring: '...', labels: [:room])
 
 # set a value
 gauge.set(21.534, labels: { room: 'kitchen' })
@@ -143,7 +143,7 @@ response sizes) and counts them in configurable buckets. It also provides a sum
 of all observed values.
 
 ```ruby
-histogram = Prometheus::Client::Histogram.new(:service_latency_seconds, docstring: '...')
+histogram = Prometheus::Client::Histogram.new(:service_latency_seconds, docstring: '...', labels: [:service])
 
 # record a value
 histogram.observe(Benchmark.realtime { service.call(arg) }, labels: { service: 'users' })
@@ -159,7 +159,7 @@ Summary, similar to histograms, is an accumulator for samples. It captures
 Numeric data and provides an efficient percentile calculation mechanism.
 
 ```ruby
-summary = Prometheus::Client::Summary.new(:service_latency_seconds, docstring: '...')
+summary = Prometheus::Client::Summary.new(:service_latency_seconds, docstring: '...', labels: [:service])
 
 # record a value
 summary.observe(Benchmark.realtime { service.call() }, labels: { service: 'database' })
@@ -168,6 +168,50 @@ summary.observe(Benchmark.realtime { service.call() }, labels: { service: 'datab
 summary.get(labels: { service: 'database' })
 # => { 0.5 => 0.1233122, 0.9 => 3.4323, 0.99 => 5.3428231 }
 ```
+
+## Labels
+
+All metrics can have labels, allowing grouping of related time series.
+
+Labels are an extremely powerful feature, but one that must be used with care.
+Refer to the best practices on [naming](https://prometheus.io/docs/practices/naming/) and 
+[labels](https://prometheus.io/docs/practices/instrumentation/#use-labels).
+
+Most importantly, avoid labels that can have a large number of possible values (high 
+cardinality). For example, an HTTP Status Code is a good label. A User ID is **not**.
+
+Labels are specified optionally when updating metrics, as a hash of `label_name => value`.
+Refer to [the Prometheus documentation](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) 
+as to what's a valid `label_name`.
+
+In order for a metric to accept labels, their names must be specified when first initializing 
+the metric. Then, when the metric is updated, all the specified labels must be present.
+
+Example:
+
+```ruby
+https_requests_total = Counter.new(:http_requests_total, docstring: '...', labels: [:service, :status_code])
+
+# increment the counter for a given label set
+https_requests_total.increment(labels: { service: "my_service", status_code: response.status_code })
+```
+
+### Pre-set Label Values
+
+You can also "pre-set" some of these label values, if they'll always be the same, so you don't
+need to specify them every time:
+
+```ruby
+https_requests_total = Counter.new(:http_requests_total, 
+                                   docstring: '...', 
+                                   labels: [:service, :status_code],
+                                   preset_labels: { service: "my_service" })
+
+# increment the counter for a given label set
+https_requests_total.increment(labels: { status_code: response.status_code })
+```
+
+
 
 ## Tests
 
--- a/README.md
+++ b/README.md
@@ -19,12 +19,12 @@ require 'prometheus/client'
 prometheus = Prometheus::Client.registry
 
 # create a new counter metric
-http_requests = Prometheus::Client::Counter.new(:http_requests, 'A counter of HTTP requests made')
+http_requests = Prometheus::Client::Counter.new(:http_requests, docstring: 'A counter of HTTP requests made')
 # register the metric
 prometheus.register(http_requests)
 
 # equivalent helper function
-http_requests = prometheus.counter(:http_requests, 'A counter of HTTP requests made')
+http_requests = prometheus.counter(:http_requests, docstring: 'A counter of HTTP requests made')
 
 # start using the counter
 http_requests.increment
@@ -99,16 +99,16 @@ The following metric types are currently supported.
 Counter is a metric that exposes merely a sum or tally of things.
 
 ```ruby
-counter = Prometheus::Client::Counter.new(:service_requests_total, '...')
+counter = Prometheus::Client::Counter.new(:service_requests_total, docstring: '...')
 
 # increment the counter for a given label set
-counter.increment({ service: 'foo' })
+counter.increment(labels: { service: 'foo' })
 
 # increment by a given value
-counter.increment({ service: 'bar' }, 5)
+counter.increment(by: 5, labels: { service: 'bar' })
 
 # get current value for a given label set
-counter.get({ service: 'bar' })
+counter.get(labels: { service: 'bar' })
 # => 5
 ```
 
@@ -118,21 +118,21 @@ Gauge is a metric that exposes merely an instantaneous value or some snapshot
 thereof.
 
 ```ruby
-gauge = Prometheus::Client::Gauge.new(:room_temperature_celsius, '...')
+gauge = Prometheus::Client::Gauge.new(:room_temperature_celsius, docstring: '...')
 
 # set a value
-gauge.set({ room: 'kitchen' }, 21.534)
+gauge.set(21.534, labels: { room: 'kitchen' })
 
 # retrieve the current value for a given label set
-gauge.get({ room: 'kitchen' })
+gauge.get(labels: { room: 'kitchen' })
 # => 21.534
 
 # increment the value (default is 1)
-gauge.increment({ room: 'kitchen' })
+gauge.increment(labels: { room: 'kitchen' })
 # => 22.534
 
 # decrement the value by a given value
-gauge.decrement({ room: 'kitchen' }, 5)
+gauge.decrement(by: 5, labels: { room: 'kitchen' })
 # => 17.534
 ```
 
@@ -143,13 +143,13 @@ response sizes) and counts them in configurable buckets. It also provides a sum
 of all observed values.
 
 ```ruby
-histogram = Prometheus::Client::Histogram.new(:service_latency_seconds, '...')
+histogram = Prometheus::Client::Histogram.new(:service_latency_seconds, docstring: '...')
 
 # record a value
-histogram.observe({ service: 'users' }, Benchmark.realtime { service.call(arg) })
+histogram.observe(Benchmark.realtime { service.call(arg) }, labels: { service: 'users' })
 
 # retrieve the current bucket values
-histogram.get({ service: 'users' })
+histogram.get(labels: { service: 'users' })
 # => { 0.005 => 3, 0.01 => 15, 0.025 => 18, ..., 2.5 => 42, 5 => 42, 10 = >42 }
 ```
 
@@ -159,13 +159,13 @@ Summary, similar to histograms, is an accumulator for samples. It captures
 Numeric data and provides an efficient percentile calculation mechanism.
 
 ```ruby
-summary = Prometheus::Client::Summary.new(:service_latency_seconds, '...')
+summary = Prometheus::Client::Summary.new(:service_latency_seconds, docstring: '...')
 
 # record a value
-summary.observe({ service: 'database' }, Benchmark.realtime { service.call() })
+summary.observe(Benchmark.realtime { service.call() }, labels: { service: 'database' })
 
 # retrieve the current quantile values
-summary.get({ service: 'database' })
+summary.get(labels: { service: 'database' })
 # => { 0.5 => 0.1233122, 0.9 => 3.4323, 0.99 => 5.3428231 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -216,6 +216,130 @@ https_requests_total.increment(labels: { status_code: response.status_code })
 
 
 
+
+## Data Stores
+
+The data for all the metrics (the internal counters associated with each labelset) 
+is stored in a global Data Store object, rather than in the metric objects themselves.
+(This "storage" is ephemeral, generally in-memory, it's not "long-term storage")
+
+The main reason to do this is that different applications may have different requirements
+for their metrics storage. Application running in pre-fork servers (like Unicorn, for
+example), require a shared store between all the processes, to be able to report coherent
+numbers. At the same time, other applications may not have this requirement but be very
+sensitive to performance, and would prefer instead a simpler, faster store.
+
+By having a standardized and simple interface that metrics use to access this store, 
+we abstract away the details of storing the data from the specific needs of each metric.
+This allows us to then simply swap around the stores based on the needs of different 
+applications, with no changes to the rest of the client. 
+
+The client provides 3 built-in stores, but if neither of these is ideal for your 
+requirements, you can easily make your own store and use that instead. More on this below.
+
+### Configuring which store to use.
+
+By default, the Client uses the `Synchronized` store, which is a simple, thread-safe Store
+for single-process scenarios.
+
+If you need to use a different store, set it in the Client Config:
+
+```ruby
+Prometheus::Client.config.data_store = Prometheus::Client::DataStores::DataStore.new(store_specific_params)
+```
+
+NOTE: You **must** make sure to set the `data_store` before initializing any metrics.
+If using Rails, you probably want to set up your Data Store on `config/application.rb`,
+or `config/environments/*`, both of which run before `config/initializers/*`
+
+Also note that `config.data_store` is set to an *instance* of a `DataStore`, not to the 
+class. This is so that the stores can receive parameters. Most of the built-in stores
+don't require any, but `DirectFileStore` does, for example.
+
+When instantiating metrics, there is an optional `store_settings` attribute. This is used
+to set up store-specific settings for each metric. For most stores, this is not used, but
+for multi-process stores, this is used to specify how to aggregate the values of each
+metric across multiple processes. For the most part, this is used for Gauges, to specify
+whether you want to report the `SUM`, `MAX` or `MIN` value observed across all processes.
+For almost all other cases, you'd leave the default (`SUM`). More on this on the 
+*Aggregation* section below.
+
+Other custom stores may also require or accept extra parameters besides `:aggregation`.
+See the documentation of each store for more details.
+
+### Built-in stores
+
+There are 3 built-in stores, with different trade-offs:
+
+- **Synchronized**: Default store. Thread safe, but not suitable for multi-process 
+  scenarios (e.g. pre-fork servers, like Unicorn). Stores data in Hashes, with all accesses
+  protected by Mutexes. 
+- **SingleThreaded**: Fastest store, but only suitable for single-threaded scenarios.
+  This store does not make any effort to synchronize access to its internal hashes, so 
+  it's absolutely not thread safe.
+- **DirectFileStore**: Stores data in binary files, one file per process and per metric.
+  This is generally the recommended store to use with pre-fork servers and other 
+  "multi-process" scenarios.
+
+  Each metric gets a file for each process, and manages its contents by storing keys and
+  binary floats next to them, and updating the offsets of those Floats directly. When 
+  exporting metrics, it will find all the files that apply to each metric, read them, 
+  and aggregate them.
+
+  In order to do this, each Metric needs an `:aggregation` setting, specifying how
+  to aggregate the multiple possible values we can get for each labelset. By default,
+  they are `SUM`med, which is what most use-cases call for (counters and histograms,
+  for example). However, for Gauges, it's possible to set `MAX` or `MIN` as aggregation, 
+  to get the highest/lowest value of all the processes / threads.
+  
+  Even though this store saves data on disk, it's still much faster than would probably be 
+  expected, because the files are never actually `fsync`ed, so the store never blocks 
+  while waiting for disk. FS caching is incredibly efficient in this regard.
+  
+  If in doubt, check the benchmark scripts described in the documentation for creating 
+  your own stores and run them in your particular runtime environment to make sure this 
+  provides adequate performance.
+
+### Building your own store, and stores other than the built-in ones.
+
+If none of these stores is suitable for your requirements, you can easily make your own.
+
+The interface and requirements of Stores are specified in detail in the `README.md`
+in the `client/data_stores` directory. This thoroughly documents how to make your own 
+store.
+
+There are also links there to non-built-in stores created by others that may be useful,
+either as they are, or as a starting point for making your own.
+
+### Aggregation settings for multi-process stores
+
+If you are in a multi-process environment (such as pre-fork servers like Unicorn), each
+process will probably keep their own counters, which need to be aggregated when receiving
+a Prometheus scrape, to report coherent total numbers.
+
+For Counters and Histograms (and quantile-less Summaries), this is simply a matter of 
+summing the values of each process.
+
+For Gauges, however, this may not be the right thing to do, depending on what they're 
+measuring. You might want to take the maximum or minimum value observed in any process,
+rather than the sum of all of them.
+
+In those cases, you should use the `store_settings` parameter when registering the 
+metric, to specify an `:aggregation` setting. 
+
+```ruby
+free_disk_space = registry.gauge(:free_disk_space_bytes,
+                                docstring: "Free disk space, in bytes",
+                                store_settings: { aggregation: :max })
+```
+
+NOTE: This will only work if the store you're using supports the `:aggregation` setting.
+Of the built-in stores, only `DirectFileStore` does.
+
+Also note that the `:aggregation` setting works for all metric types, not just for gauges. 
+It would be unusual to use it for anything other than gauges, but if your use-case 
+requires it, the store will respect your aggregation wishes.
+
 ## Tests
 
 Install necessary development gems with `bundle install` and run tests with
--- a/examples/rack/README.md
+++ b/examples/rack/README.md
@@ -49,6 +49,8 @@ something like this:
 
 ```ruby
 use Prometheus::Middleware::Collector, counter_label_builder: ->(env, code) {
+  next { code: nil, method: nil, host: nil, path: nil } if env.empty?
+
   {
     code:         code,
     method:       env['REQUEST_METHOD'].downcase,
--- a/lib/prometheus/client/data_stores/README.md
+++ b/lib/prometheus/client/data_stores/README.md
@@ -195,11 +195,11 @@ repository for these.
 This is just an example of how one could implement a data store, and a clarification on
 the "aggregation" point 
 
-Important: This is **VAPORWARE**, intended simply to show how this could work / how to
+Important: This is a **toy example**, intended simply to show how this could work / how to
 implement these interfaces.
 
 There are some key pieces of code missing, which are fairly uninteresting, this only shows
-the parts that illustrate the idea of storing multiple different values, and aggregate
+the parts that illustrate the idea of storing multiple different values, and aggregating
 them
 
 ```ruby
--- a/lib/prometheus/client/data_stores/README.md
+++ b/lib/prometheus/client/data_stores/README.md
@@ -1,0 +1,285 @@
+# Custom Data Stores
+
+Stores are basically an abstraction over a Hash, whose keys are in turn a Hash of labels
+plus a metric name. The intention behind having different data stores is solving 
+different requirements for different production scenarios, or performance trade-offs.
+
+The most common of these scenarios are pre-fork servers like Unicorn, which have multiple
+separate processes gathering metrics. If each of these had their own store, the metrics
+reported on each Prometheus scrape would be different, depending on which process handles
+the request. Solving this requires some sort of shared storage between these processes, 
+and there are many ways to solve this problem, each with their own trade-offs.
+
+This abstraction allows us to easily plug in the most adequate store for each scenario.
+
+## Interface
+   
+`Store` exposes a `for_metric` method, which returns a store-specific and metric-specific 
+`MetricStore` object, which represents a "view" onto the actual internal storage for one
+particular metric. Each metric / collector object will have a references to this 
+`MetricStore` and interact with it directly. 
+
+The `MetricStore` class must expose `synchronize`, `set`, `increment`, `get` and `all_values`
+methods,  which are explained in the code sample below. Its initializer should be called 
+only by `Store#for_metric`, not directly.
+
+All values stored are `Float`s.
+
+Internally, a `Store` can store the data however it needs to, based on its requirements. 
+For example, a store that needs to work in a multi-process environment needs to have a 
+shared section of memory, via either Files, an MMap, an external database, or whatever the 
+implementor chooses for their particular use case.
+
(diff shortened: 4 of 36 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ before_install:
   - |
     if [[ "$(ruby -e 'puts RUBY_VERSION')" != 1.* ]]; then gem update --system; fi
 rvm:
-  - 1.9.3
-  - 2.3.3
-  - 2.4.0
+  - 2.3.8
+  - 2.4.5
+  - 2.5.3
```
lzap: Just a note - Ruby 2.0 is EOL in upstream, but it's still part as Red Hat Enterprise Linux 7 and all its clones with guaranteed support of 10+ years. If there is any chance of keeping 2.0.0 it's not a bad idea to keep it for another few years (2024 is EOL for Ruby 2.0 in CentOS7, a bit later for RHEL7 extended support).

What I am trying to say, there is a lot of software running just fine on Ruby 2.0.x and if there is not strong reason to ditch it, I'd vote keeping it.

lzap:
Well done, this is not a review but I like the idea, the store API looks great. Thanks.

.travis.yml
dmagliola: Thank you for your feedback. Some thoughts on this:
- 2.0 doesn't support required keyword arguments, so we'd need at least 2.1. We could side-step this by not using them, but I much prefer how the interface feels with them.
- Our test suite passes in Ruby 2.1.10. So, while we're not running CI on it, you should be able to use this in 2.1

lib/prometheus/client/data_stores/README.md
```diff
@@ -0,0 +1,286 @@
...
+`it_behaves_like Prometheus::Client::DataStores`
+
+Follow the simple structure in `synchronized_spec.rb` for a starting point.
+
+Note that if your store stores data somewhere other than in-memory (in files, Redis, 
+databases, etc), you will need to do cleanup between tests in a `before` block.
+
+The tests for `DirectFileStore` have a good example at the top of the file. This file also
+has some examples on testing multi-process stores, checking that aggregation between 
+processes works correctly.
```
lzap: It's tempting to say aggregation should be a separate concern, but some stores actually aggregate incoming data. So your design is correct.

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -0,0 +1,302 @@
...
+      # When exporting metrics, the process that gets scraped by Prometheus  will find
+      # all the files that apply to a metric, read their contents, and aggregate them
+      # (generally that means SUMming the values for each labelset).
+      #
+      # In order to do this, each Metric needs an `:aggregation` setting, specifying how
+      # to aggregate the multiple possible values we can get for each labelset. By default,
+      # they are `SUM`med, which is what most use cases call for (counters and histograms,
+      # for example).
+      # However, for Gauges, it's possible to set `MAX` or `MIN` as aggregation, to get
+      # the highest value of all the processes / threads.
```
lzap: Out of curiosity, have you considered Redis?

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -0,0 +1,302 @@
...
+                  # "foo=bar&x=y" => { "foo" => ["bar"], "x" => ["y"] }
+                  # Turn the keys back into symbols, and remove the arrays
+                  label_set = CGI::parse(labelset_qs).map do |k, vs|
+                    [k.to_sym, vs.first]
+                  end.to_h
+
+                  stores_data[label_set] << v
+                end
+              end
+              store.close
```
lzap: Probably `ensure`?

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -0,0 +1,302 @@
...
+          end
+
+          # Filename for this metric's PStore (one per process)
+          def filemap_filename
+            filename = "metric_#{ metric_name }___#{ process_id }.bin"
+            File.join(@store_settings[:dir], filename)
+          end
+
+          def stores_for_metric
+            Dir.glob(File.join(@store_settings[:dir], "metric_#{ metric_name }___*"))
```
lzap: Nicely done.

lib/prometheus/client/data_stores/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+          # Metrics may need to modify multiple values at once (Histograms do this, for 
+          # example). MetricStore needs to provide a way to synchronize those, in addition
+          # to all of the value modifications being thread-safe without a need for simple 
+          # Metrics to call `synchronize`
+          def synchronize
+            raise NotImplementedError
+          end
+
+
```
lzap: Extra line.

lzap:
Benchmark results from my AMD Ryzen 1700 (SMT turned off):

```
[lzap@box client_ruby]$ bundle exec ruby spec/benchmarks/labels.rb
Warming up --------------------------------------
            0 labels   198.881k i/100ms
            2 labels    42.070k i/100ms
          100 labels     1.608k i/100ms
  2 lab, half cached    42.670k i/100ms
100 lab, half cached     1.677k i/100ms
   2 lab, all cached   198.826k i/100ms
 100 lab, all cached   198.719k i/100ms
Calculating -------------------------------------
            0 labels      3.353M (± 0.4%) i/s -     16.905M in   5.042115s
            2 labels    485.273k (± 0.7%) i/s -      2.440M in   5.028458s
          100 labels     16.255k (± 1.7%) i/s -     82.008k in   5.046815s
  2 lab, half cached    489.152k (± 0.9%) i/s -      2.475M in   5.059884s
100 lab, half cached     16.881k (± 1.7%) i/s -     85.527k in   5.068145s
   2 lab, all cached      3.330M (± 1.3%) i/s -     16.701M in   5.017013s
 100 lab, all cached      3.341M (± 0.6%) i/s -     16.891M in   5.055868s
[lzap@box client_ruby]$ bundle exec ruby spec/benchmarks/data_stores.rb
                                                    user     system      total        real
Observe NoopStore                 x1            1.036507   0.011970   1.048477 (  1.053327)
Export  NoopStore                 x1            0.000779   0.000000   0.000779 (  0.000787)
Observe SingleThreaded            x1            5.407055   0.009963   5.417018 (  5.438107)
Export  SingleThreaded            x1            0.012109   0.000000   0.012109 (  0.012442)
Observe Synchronized              x1           11.390798   0.001001  11.391799 ( 11.437172)
Export  Synchronized              x1            0.010183   0.000001   0.010184 (  0.010500)
Observe DirectFileStore           x1           15.236973   1.937662  17.174635 ( 17.257330)
Export  DirectFileStore           x1            0.025676   0.010973   0.036649 (  0.037361)
--------------------------------------------------------------------------------
Observe NoopStore                 x2            1.242889   0.022793   1.265682 (  1.265861)
Export  NoopStore                 x2            0.000980   0.000000   0.000980 (  0.000990)
Observe Synchronized              x2           11.727178   0.005978  11.733156 ( 11.782778)
Export  Synchronized              x2            0.006566   0.000000   0.006566 (  0.006615)
Observe DirectFileStore           x2           28.208413   7.853636  36.062049 ( 27.762403)
Export  DirectFileStore           x2            0.022053   0.004099   0.026152 (  0.026341)
--------------------------------------------------------------------------------
Observe NoopStore                 x4            1.297687   0.000000   1.297687 (  1.302745)
Export  NoopStore                 x4            0.001272   0.000000   0.001272 (  0.001279)
Observe Synchronized              x4           11.939564   0.000000  11.939564 ( 11.988349)
Export  Synchronized              x4            0.011443   0.000000   0.011443 (  0.011604)
Observe DirectFileStore           x4           33.324043  13.822482  47.146525 ( 34.224238)
Export  DirectFileStore           x4            0.023723   0.008286   0.032009 (  0.032313)
--------------------------------------------------------------------------------
Observe NoopStore                 x8            1.108902   0.000000   1.108902 (  1.113839)
Export  NoopStore                 x8            0.002330   0.000000   0.002330 (  0.002361)
Observe Synchronized              x8           12.362316   0.000000  12.362316 ( 12.434820)
Export  Synchronized              x8            0.013203   0.000000   0.013203 (  0.013319)
Observe DirectFileStore           x8           36.536457  17.406428  53.942885 ( 39.150231)
Export  DirectFileStore           x8            0.014251   0.009418   0.023669 (  0.023827)
--------------------------------------------------------------------------------
Observe NoopStore                 x12           1.503590   0.000000   1.503590 (  1.509897)
Export  NoopStore                 x12           0.001449   0.000000   0.001449 (  0.001467)
Observe Synchronized              x12          12.534795   0.000000  12.534795 ( 12.610278)
Export  Synchronized              x12           0.014397   0.000000   0.014397 (  0.014517)
Observe DirectFileStore           x12          41.475597  19.995821  61.471418 ( 47.487212)
Export  DirectFileStore           x12           0.021214   0.005016   0.026230 (  0.026409)
--------------------------------------------------------------------------------
Observe NoopStore                 x16           1.612725   0.000000   1.612725 (  1.620807)
Export  NoopStore                 x16           0.001397   0.000000   0.001397 (  0.001341)
Observe Synchronized              x16          12.510221   0.000000  12.510221 ( 12.580589)
Export  Synchronized              x16           0.013395   0.000000   0.013395 (  0.013541)
Observe DirectFileStore           x16          42.696869  21.355741  64.052610 ( 49.881440)
Export  DirectFileStore           x16           0.018934   0.007126   0.026060 (  0.026312)
--------------------------------------------------------------------------------
Observe NoopStore                 x20           1.646399   0.000000   1.646399 (  1.657064)
Export  NoopStore                 x20           0.001430   0.000000   0.001430 (  0.001438)
Observe Synchronized              x20          12.962815   0.000000  12.962815 ( 13.058244)
Export  Synchronized              x20           0.012914   0.000000   0.012914 (  0.013018)
Observe DirectFileStore           x20          43.319909  21.250377  64.570286 ( 50.071546)
Export  DirectFileStore           x20           0.021107   0.004597   0.025704 (  0.026774)
--------------------------------------------------------------------------------
```

lzap:
@grobie or anyone? What is the status? This is bunch of great work right there.

lib/prometheus/client/data_stores/direct_file_store.rb
dmagliola: I have. And from a personal point of view, I really wanted it to be the best (I love Redis). Performance is a lot worse than the other options, though. About 65 microseconds for each call, compared to 9 for the DirectFileStore

spec/benchmarks/README.md
```diff
@@ -0,0 +1,65 @@
...
+  want to have fewer labels instead.
+  
+It should be easy to modify the constants in this benchmark to your particular situation,
+if necessary.
+
+## Running the benchmarks
+
+Simply run, from the repo's root directory:
+
+`bundle exec ruby spec/benchmarks/benchmark_file.rb`
```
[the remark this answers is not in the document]
dmagliola: Right, I thought that was self-explanatory, and catered for adding more benchmarks in the future, but I like your suggestion

README.md
```diff
@@ -19,12 +19,12 @@ require 'prometheus/client'
 prometheus = Prometheus::Client.registry
 
 # create a new counter metric
-http_requests = Prometheus::Client::Counter.new(:http_requests, 'A counter of HTTP requests made')
+http_requests = Prometheus::Client::Counter.new(:http_requests, docstring: 'A counter of HTTP requests made')
```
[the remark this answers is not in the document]
dmagliola: I'm generally with you on the "not breaking the contract". This is, for example, why I didn't change the namespaces, even though I'd prefer a "flatter" namespace structure, and i've seen comments around about that too (can't find one right now, though).
That seemed like "unnecessarily" breaking the contract.

However, we're breaking the contract on other places, mainly, demanding that labels be declared up-front when a new metric is instantiated. I consider that a "good" break of the contract, since it takes us more in line with best practices.

`docstring` is the one i'm most on the fence about.
I don't fully like it, but I also don't like the two nakes strings (metric_name and docstring). I feel like the first one is pretty self-explanatory, but the second one not so much.

That said, if you want me to make docstring a non-kwarg, I'll happily do so.

README.md
```diff
@@ -169,6 +169,50 @@ summary.get(labels: { service: 'database' })
...
+                                   docstring: '...', 
+                                   labels: [:service, :status_code],
+                                   preset_labels: { service: "my_service" })
+
+# increment the counter for a given label set
+https_requests_total.increment(labels: { status_code: response.status_code })
+```
+
+
+
```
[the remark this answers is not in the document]
dmagliola: It would *certainly* be a major version bump...
If we were above `1.0` :)
Semver says that below 1.0, anything goes (or at least, that's my understanding, I may be wrong)

But yes, this is definitely not backwards compatible. I've tried to keep it as compatible as possible, but there's no way to keep backwards compatibility *and* follow the current best practices.
That said, I've tried to minimize pain for existing consumers, while also making it better for future ones. 
Best practices suggest the method be called `inc`, for example. I chose to leave it as `increment`, because it felt like a needless contract break. But i also change the signature because the new one is better for the case we're supposed to cater the most for, which is no labels.

lib/prometheus/middleware/collector.rb
```diff
@@ -20,6 +20,11 @@ module Middleware
     #
     # The request duration metric is broken down by method and path by default.
     # Set the `:duration_label_builder` option to use a custom label builder.
+    #
+    # Label Builder functions will receive a Rack env and a status code, and must
+    # return a hash with the labels for that request. They must also accept an empty
+    # env, and return a hash with the correct keys. This is necessary to initialize
+    # the metrics with the correct set of labels.
```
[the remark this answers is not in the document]
dmagliola: :)

README.md
```diff
@@ -158,15 +158,18 @@ histogram.get(labels: { service: 'users' })
 Summary, similar to histograms, is an accumulator for samples. It captures
 Numeric data and provides an efficient percentile calculation mechanism.
 
+For now, only `sum` and `total` (count of observations) are supported, no actual quantiles.
```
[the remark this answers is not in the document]
dmagliola: Let's look at that, i've added it to my list. But I propose doing that as a separate PR, this one is big enough already

lib/prometheus/client/data_stores/direct_file_store.rb
dmagliola: As an addition to that comment, as I've mentioned in a few places, we're making a second repo (it's empty for now, sorry, will be up next week)
https://github.com/gocardless/prometheus-client-ruby-data-stores-experiments

In there we'll have all the stores we tried, plus all the experiments, benchmark results, etc, etc. Basically a lot of documentation for people planning to make their own stores. So we can save time for someone that might try Redis, for example, since we did that already :)

This PR (and thus this repo) only has the "winners" from all the experiments, what we consider the best trade-offs.

lib/prometheus/client/data_stores/direct_file_store.rb
dmagliola: Yes, good catch@

lib/prometheus/client/data_stores/direct_file_store.rb
dmagliola: :)

dmagliola:
@lzap Thanks SO much for the review!
I've added a couple `fixup` commits rather than rebase, so all the comments don't get hidden. Once this is ready to merge will rebase those, and "re-signoff"
I hope I've responded to all your comments. Let me know if you have any other feedback!

-----------------------------------
As for your benchmark:
That looks *roughly* within the same ballpark as our numbers. At least, the stores have roughly the same proportion to each other.
I was planning to publish this info in the `data-stores-experiments` repo, which is why I left it off here, but this is the machine our benchmarks ran on:

GCE (Google Cloud) instance with:
  - 4 vCPUs, 4 GB memory
  - Intel Skylake
  - standard persistent disk (not SSD)
  - europe-west4-a
  - the directory that the `DirectFileStore` was writing to is mounted in `tmpfs`.

This last bit doesn't make as much difference as one would expect (again, because FS caching):
- SSD: 10.241138
- TmpFS: 9.147389

So, TmpFS is about 10% faster, which is interesting, but not that world-shattering.
Also, HDD and SSD seem to be about the same (again, because of FS caching, and no `fsyncing`)

spec/benchmarks/README.md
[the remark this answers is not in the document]
lzap: Oh now I understand jeez I blindly copy pasted it. Uh.

README.md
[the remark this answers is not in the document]
lzap: I figured out later that API is changing, no big deal. It is used widely in the library anyway.

examples/rack/README.md
```diff
@@ -49,6 +49,8 @@ something like this:
 
 ```ruby
 use Prometheus::Middleware::Collector, counter_label_builder: ->(env, code) {
+  next { code: nil, method: nil, host: nil, path: nil } if env.empty?
```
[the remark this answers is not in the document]
lzap: That is probably for more experienced Prom guys, but honestly the first thing I wrote in our "telemetry API" Prom wrapper was up-front label definition for easier auto-mapping for statsd. This confirms your idea, if you are curious what we do: https://github.com/theforeman/foreman/pull/5096

lib/prometheus/client/data_stores/direct_file_store.rb
lzap: Wow, this is exceptional contribution, you really made your homework.

Now, it is somewhat surprising to me that Redis is slower than DirectFileStore (I am assuming on tmpfs). It has quite pricey communication channel (text over TCP), but on the other hand it is an in-memory store. Interesting result, would not expect to see DirectFileStore as the winner!

This brings one remark. Can you share some recommendation (in one of the README files) about filesystem choice. My dumb assumption is that it only makes sense with tmpfs for the fastest possible performance (data-loss is not a thing here). Or am I missing something and you are actually experimenting with normal FS like ext4 or xfs?

lzap:
To me it looks like the best advice would be:

* Use tmpfs if you don't care about data-loss, have at least 1GB of swap so unused values can be dropped off.
* Use ext4 preferably with dir_index option (default these days in most distros) on a separate partition with fsync set to 30 seconds or something like that for semi-safe setup.

No problem, I don't have push permissions here just passing-by so don't get excited. Real experts need to do their assessment, but I am willing to help with testing once this is in RTM state. Will definitely take a look how it performs for our app.

lib/prometheus/client/data_stores/direct_file_store.rb
dmagliola: I found it very surprising too, however, I believe it makes sense. 
Redis is very fast compared to disk-based databases... But disks are only slow if you care about your data staying there. If you don't (and in this unusual case, we don't), you don't need to synchronously hit the slow disk. The FS will cache stuff for you in memory, and flush to disk whenever it wants in the background.

This is also why the FS matters less than one would expect. I did run my benchmarks on TmpFS, and it's about 10% faster than a normal filesystem, which is a lot less speedup than I would've expected. The numbers are also almost the same for example in an SSD or an HDD, which sounds even more counter-intuitive, but we're not really hitting the disk much, so it doesn't really matter that much what's behind it.

grobie:
I'm so sorry for the delay @lzap. Thank you so much for the great work. I haven't had much time to spend on open-source projects during the last months unfortunately.

We're wrapping up our projects we had to deliver this year and will have time next week for a proper review of this pr.

lzap:
Heh thanks for the update, this is all @dmagliola work I am just happen to be random visitor. But if you need more assistance with testing this, I am more than happy to try this out.

dmagliola:
Hey @grobie!

Glad to hear you'll have time to look into this soon!

As a reminder, it's worth looking at this PR commit-by-commit. Looking at the whole diff as one unit isn't very practical. I've tried to organise the commits in a way that's easy to follow, with well explained steps. The long commit messages document the intent of every step.

Also, if you want to discuss this, I'm available for whatever you need. If you find you have questions that would be better served by higher-bandwidth discussion, we can hop on a Skype call and drop any decisions made there back into this thread.

Thanks!

stickyd:
Hello! Is there anything I can do to help get this through? Very important to us :)

lzap:
We are eager to see this at Red Hat too. Offering assistance of any kind in this, testing, co-maintaining the project, whatever is needed.

Sinjo:
Hi @grobie 👋🏼

I'm one of @dmagliola's colleagues, and I've got some good news!

We've got this branch running in production, and we're using the `DirectFileStore` in our multi-process Unicorn web servers. For the first time, we've got Prometheus scraping metrics from our web tier, and things look good so far.

We're keen to offer whatever help we can to get the work upstreamed. As a primarily Ruby shop that's betting heavily on Prometheus for monitoring, we're also happy to stick around in the long run and help however we can.

Let me know if there's anything we can do.

Cheers!

grobie:
Thanks @lzap @Sinjo @dmagliola for offering help with the maintenance of this project, and apologies for my radio silence. As you have surely already guessed, I don't have the time anymore to maintain this client library.

I'd be very much interested in someone taking over as maintainer of this project. If you're still willing to do this, please write me an email to tobidt@gmail.com and let me know your timezone. I'll then set up a quick call with you to discuss this in person.

lzap:
Thanks for the update, gonna drop you a line. I am all for having multiple maintainers, let's build a health community around this nice project!

cemeng:
:+1: this is a great effort, hoping to see this request progresses - really keen on using it on our project (have played around with the branch on our app locally - seems to work great with unicorn).

dmagliola:
Update: Pushed a new set of commits to resolve the conflict with `master`

README.md
```diff
@@ -158,17 +158,228 @@ histogram.get({ service: 'users' })
...
-# retrieve the current quantile values
-summary.get({ service: 'database' })
-# => { 0.5 => 0.1233122, 0.9 => 3.4323, 0.99 => 5.3428231 }
+# increment the counter for a given label set
+https_requests_total.increment(labels: { service: "my_service", status_code: response.status_code })
+```
+
+### Pre-set Label Values
+
+You can also "pre-set" some of these label values, if they'll always be the same, so you don't
```
brian-brazil: We've found that this tends to get misunderstood/abused for target labels, I'd suggest not having this.

README.md
```diff
@@ -158,17 +158,228 @@ histogram.get({ service: 'users' })
...
+lead to performance improvements. If you are incrementing a counter in a fast loop, you
+definitely want to be doing this.
+
+
+Examples:
+
+**Pre-setting labels for ease of use:**
+
+```ruby
+# in the file where you define your metrics:
```
brian-brazil: Metric definitions should almost always be in the file they're used.

README.md
```diff
@@ -158,17 +158,228 @@ histogram.get({ service: 'users' })
...
+
+  In order to do this, each Metric needs an `:aggregation` setting, specifying how
+  to aggregate the multiple possible values we can get for each labelset. By default,
+  they are `SUM`med, which is what most use-cases call for (counters and histograms,
+  for example). However, for Gauges, it's possible to set `MAX` or `MIN` as aggregation, 
+  to get the highest/lowest value of all the processes / threads.
+  
+  Even though this store saves data on disk, it's still much faster than would probably be 
+  expected, because the files are never actually `fsync`ed, so the store never blocks 
+  while waiting for disk. FS caching is incredibly efficient in this regard.
```
brian-brazil: Technically the kernel's page cache, not the filesystem

README.md
```diff
@@ -158,17 +158,228 @@ histogram.get({ service: 'users' })
...
+free_disk_space = registry.gauge(:free_disk_space_bytes,
+                                docstring: "Free disk space, in bytes",
+                                store_settings: { aggregation: :max })
+```
+
+NOTE: This will only work if the store you're using supports the `:aggregation` setting.
+Of the built-in stores, only `DirectFileStore` does.
+
+Also note that the `:aggregation` setting works for all metric types, not just for gauges. 
+It would be unusual to use it for anything other than gauges, but if your use-case
```
brian-brazil: I don't think it makes sense for this to be available for anything other than gauges. Is there a use case?

lib/prometheus/client/data_stores/README.md
```diff
@@ -0,0 +1,306 @@
...
+## Conventions
+
+- Your store MAY require or accept extra settings for each metric on the call to `for_metric`.
+- You SHOULD validate these parameters to make sure they are correct, and raise if they aren't. 
+- If your store needs to aggregate multiple values for the same metric (for example, in
+  a multi-process scenario), you MUST accept a setting to define how values are aggregated.
+  - This setting MUST be called `:aggregation`
+  - It MUST support, at least, `:sum`, `:max` and `:min`. 
+  - It MAY support other aggregation modes that may apply to your requirements.
+  - It MUST default to `:sum`
```
brian-brazil: This should be consistent with the python client for sanity, which is to expose each individually.

lib/prometheus/client/data_stores/README.md
```diff
@@ -0,0 +1,306 @@
...
+        end
+      end
+    end
+  end
+end
+```
+
+## Conventions
+
+- Your store MAY require or accept extra settings for each metric on the call to `for_metric`.
```
brian-brazil: It shouldn't ever require, as that'd mean that every metric in every library you use would need to be aware of it.

errm:
Overall this looks great. We are currently using the gitlab fork... but I would love to switch to a more official client library...

I might be missing something but, as far as I can tell there doesn't seem to be a way to to expose the metrics from each process individually ... I think this should at least be possible ... if not the default mode of operation.

lib/prometheus/client/data_stores/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+Check out the [GoCardless Data Stores Experiments](gocardless/prometheus-client-ruby-data-stores-experiments) 
+repository for these.
+
+## Sample, imaginary multi-process Data Store
+
+This is just an example of how one could implement a data store, and a clarification on
+the "aggregation" point 
+
+Important: This is **VAPORWARE**, intended simply to show how this could work / how to
```
errm: I am not sure that **VAPORWARE** is quite the right description for this... see https://en.wikipedia.org/wiki/Vaporware

Perhaps some other description would be better, some suggestions:
* contrived example
* toy example
* illustrative code

Sinjo: Agreed, and changed.

README.md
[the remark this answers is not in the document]
Sinjo: Taking note of the idea of an upgrading section, and marking this thread as resolved.

Sinjo:
Hey everyone,

I realise we've gone quiet on this PR, so I want to give you an update on where we're up to and what we plan to do next.

Myself, @dmagliola, and @SuperQ [took on maintainership](https://github.com/prometheus/client_ruby/pull/103) of this gem at beginning of the month.

We (GoCardless, where the PR was written) have been running this branch - with the `DirectFileStore` backend in particular - in production for almost 2 months now. It's worked as expected, and given us aggregated metrics from our forking web server backends.

We're keen to get this work into a stable release and into the hands of the wider community. Before we can do that, we need to address the outstanding feedback we've received on this PR, as well as feedback we've received from our internal users.

My plan is to draw up a checklist of what's left to do to get this PR ready to merge and work through it. I'll add that checklist to the top of the PR so that people can see how it's progressing.

I'm keen not to make this PR much bigger than it already is, so for any enhancements, extra docs, and changes that aren't fundamental to the approach we're taking, I'm going to spin out separate GitHub issues and attach them to a new release milestone.

Thanks to everyone who's been involved, and hopefully we'll have something ready for the community soon.

Cheers!

SuperQ:
@Sinjo Thanks, a checklist sounds like a good idea. I was also going to invite @stanhu and @splattael to take a look over this from the GitLab side.

Sinjo:
@SuperQ Sorry for the radio silence - busy week at work with stuff unrelated to Prometheus. That review would be awesome!

I'm just collating the list now. Hopefully have it up some time tomorrow.

SuperQ:
I've added an internal deliverable for providing a review of this. https://gitlab.com/gitlab-org/gitlab-ce/issues/59959

README.md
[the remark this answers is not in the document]
Sinjo: Resolving this as we're not going to block this work on summary quantile support.

Sinjo:
Right, so I've been through the PR and the feedback we got internally looking for what we should do next, and I've ended up writing quite the list!

@SuperQ: Given the length of this list, my preference is to merge this PR, create issues for everything we want to do, attach them to a milestone, and work through them in individual PRs. Already discussed the idea with @dmagliola and he's up for that approach. Does it sound alright to you?

# Definitely before release

These items are things we should do before we cut a release of the gem. In some cases we just need to make a decision one way or the other or write some documentation. In others we need to change some code.

## Memory usage of `DirectFileStore`

From running it in production, we've found `DirectFileStore` to be more memory-hungry than other ways of storing metrics.

This isn't entirely surprising, but we've done a little investigation into whether we could reduce the effect and we may have some improvements we can make.

At a very minimum, we should document that it uses more memory than the hash-based stores.

## Consider exporting unaggregated per-pid metrics by default

We had a [couple of](https://github.com/prometheus/client_ruby/pull/95#discussion_r262055780) [comments](https://github.com/prometheus/client_ruby/pull/95#issuecomment-471575689) on this point, and it’s fair. The point about “no obvious right way to aggregate gauges in all cases” is a solid argument.

We need to decide a stance on this both for gauges specifically and for metrics in general (i.e. should gauges be a special case where we don't aggregate by default).

## Review approach to file globbing when reading metrics

One of our internal users raised a point about our method of reading PID files when exporting metrics making it possible to accidentally include more files than it should. Specifically, if one metric name is a subset of another, the export of the metric with the subset name could include values from the metric with a longer name.

@dmagliola commented that running into this issue would involve putting a triple underscore in your metric name, which is highly unusual and against conventions, but maybe we can choose a character that never appears in metric names when we generate the file names.

At a minimum, if we make no code change, we should document the behaviour.

## Validate behaviour of `DirectFileStore` in the presence of existing files

We (GoCardless) run our services in containers, which means a clean file system every time we boot the app.

We should look at what the behaviour is like for people who have file systems that persist between versions of the app. If so, we should look at what mitigations we can implement to make `DirectFileStore` work by default.

Any edge-cases should be added to the `DirectFileStore`’s docs.

## Resolve discussion on Ruby version support

I'm pretty strongly against supporting anything below Ruby 2.1 because the lack of required keyword arguments is a pain to work around (you can, with sentinel values, but it's a mess). It's been out of support for a long time now that I don't think it justifies the ongoing effort and risk of bugs in our workarounds.

This also raises the discussion of what our Ruby version support policy should be overall. I think it's good to document this up front, to set expectations around how we'll act as maintainers - something we've [done before on our own open source projects](https://github.com/gocardless/statesman/blob/fdab116b9b0d9e1d77b5df151641761fbc7c15dd/docs/COMPATIBILITY.md).

Tangientially, some of the memory optimisations people have been playing with here involve methods introduced to Ruby's stdlib in relatively new versions (one of them was added in 2.5). If they're really worth having, we might be looking at some code that conditionally runs in those versions.

## Decide if Rack middleware should support flexible labels

Doing this commits us to maintaining what is currently quite a [confusing API](https://github.com/prometheus/client_ruby/pull/95#discussion_r233501944).

We can always add this functionality later, but it will be a pain to take it away.

I’m in favour of only supporting our fixed set of labels in the Rack middleware we provide, and having a README section advising people to do their own thing if they want something more sophisticated.

## Resolve discussion on pre-set label values

These have been [pointed out](https://github.com/prometheus/client_ruby/pull/95#discussion_r262053645) as a feature that gets abused.

Right now I’m not exactly sure what that abuse typically looks like, but maybe there are Prometheus docs that talk about it.

Is the abuse so bad that clients are being built without this feature now?

## Minor tweaks to README based on @brian-brazil’s feedback

Some easy fixes that we should just make (e.g. [filesystem cache -> kernel page cache](https://github.com/prometheus/client_ruby/pull/95#discussion_r262054398)).

## Resolve discussion on required settings in `for_metric`

Another piece of [PR feedback](https://github.com/prometheus/client_ruby/pull/95#discussion_r262056166). It's not something we use ourselves. It was there to give flexibility to future stores people might write.

I can see how it would play badly with pre-canned metrics exported by libraries. Let's take it out of the docs.

## Types of stored labels (symbols vs strings)

When you use `DirectFileStore`, because of the nature of the store, your label values will always come back as strings. In other stores if you pass them in as symbols they'll come back as symbols.

This is a small gotcha that you'll only notice if you swap between store implementations and read metric data out of the store yourself (i.e. not through the formatter). It's most likely that you'll notice this in your tests.

We could:
  - Change nothing
  - Require that all stores convert label values to strings

## Initial value of initialised but unset metrics (`nil` vs `0.0`)

Our data store interface is defined as a map of strings to strings to floats. Our in-memory stores allow for `nil` as well (when the value hasn't been initialised). `DirectFileStore` doesn't, because it only knows how to round-trip a float for a given metric/label combination.

Again, it's most likely that you'll notice this in your tests if you try to read back a metric value.

## Add CI matrix for all supported Ruby versions

Hygiene. We should just do it.

## Documentation for upgrading

This is a good idea, given how much change we're making Already [moved out](https://github.com/prometheus/client_ruby/issues/104) to its own issue.

## Documentation for `DirectFileStore`

We've got a few mentions of it in `README.md` and in the stores documentation, but I think we'd do well to have specific documentation for this store, particularly around any usage constraints/requirements. I'd expect this to cover at least the caveats mentioned above if we can't solve them more cleanly.

# Deferrable until after release

These items are worth recording issues for, but aren't blockers to a 1.0 release.

## Small script in bin to dump contents of DirectFileStore files

A [good idea](https://github.com/prometheus/client_ruby/pull/95#discussion_r233515841) for a debugging utility the would help when working on `DirectFileStore`.

Sinjo:
@SuperQ Any thoughts on merging and spinning that list out to separate PRs? If we go that route, we're planning to get the team I'm in here to make a big push to work through them.

SuperQ:
Sorry for the delay in reviewing your previous update.

I am in favor of merging this, and moving forward with additional changes / discussions in new issues/PRs.

Maybe we should take all of the discussion items you posted and create issues for them.

Sinjo:
Awesome. I've gone ahead and merged, and I'll spin the list out into a bunch of separate issues.

philandstuff:
Can I request a new release of current master?  We're already using this branch in alphagov/verify-frontend, and now that this has been merged, I can adjust our Gemfile to point at the prometheus/client_ruby instead of gocardless/prometheus_client_ruby; but it'd be nicer still to use a `0.10.0` or a `1.0` release from rubygems.

Sinjo:
Alright. Issues moved out and linked from the list above. [Milestone](https://github.com/prometheus/client_ruby/milestone/3) created for everything we want to do pre-1.0.

There's a chance we can make a smaller 0.10.0 milestone with only the most egregious breaking changes in it. I'll weigh it up based on how quickly we can churn through the list.

I think we're done with this mega-PR. 😅

lzap:
Great talk, well done! :-)

dmagliola:
Thank you @lzap !!

lib/prometheus/client/data_stores/direct_file_store.rb
liaden: Would this be worth revisiting redis as an option but with https://github.com/socketry/async-redis? I was able to update the original Redis client with changes that have since been adopted for `DirectFileStore` to make it work (but obviously still slower than `DirectFileStore`).

dmagliola: Maybe, but what would be the objective?
Is the idea that it'd be faster than DirectFileStore?

Because in my mind that approach has the very clear downside that now you need a local Redis server running on each of your servers/pods/etc (you should ideally not use a central Redis for your metrics), and we introduce `async-redis` as a dependency to this gem.

What's the trade-off we're looking for?

liaden: My hope was that it would reduce performance penalty to being closer to in line with DirectFileStore which could allow for running a central redis for metrics or running a localhost redis.

Having to run redis locally is a downside, but it comes with the perk of running a containered ruby application with a read only filesystem. That said, I can make the DirectFileStore work with a TmpFS volume for the container and still maintain a read only filesystem. With DirectFileStore being memory mounted, I do have to account for that with setting memory limits on my ruby container. With nomad or kubernetes, I could configure a redis container as a sidecar which makes memory impacts a bit easier to track.

If async-redis is reasonable with a centralized redis, then the sidekiq redis can be reused (with another DB to not overlap with sidekiq), thus not increasing the operational burden.

That said, the downsides of redis may be enough that it isn't worth adding to this gem and that is why I wanted to double check before I tried adapting for async.

dmagliola: Interesting, never thought about the read-only FS use case. @Sinjo FYI :)

I don't think we want to ship a Redis data store as part of the official client, however, it's easy to make your own.
Here's one that already works: https://github.com/gocardless/prometheus-client-ruby-data-stores-experiments/blob/master/data_stores/redis.rb
(or did when I made it, I don't remember if the Stores interface changed since then, but it should be at least easy to adapt)

That one is not async, but you can use it as a starting point.

As for using a centralized Redis... This is generally not considered a great practice, as it'll combine the metrics from all your servers into one. If you have *one* server that is misbehaving, you would not be able to see that. This is why it's generally recommended to run a local Redis on each one of your servers.

Sinjo: > I don't think we want to ship a Redis data store as part of the official client

I'd agree on this point. I don't think a Redis data store offers significant enough benefits to include it with the client librarly. `DirectFileStore` pretty thoroughly covers the use case of multi-process web servers, and my first thought for read-only filesystems is tmpfs (as you mentioned).

I'd echo what Daniel said about using a centralised Redis instance, and add another reason: putting your metric storage across a network hop could get weird. In theory it should be fine if all the operations are async, but it's something I'd personally prefer not to have to think about if I could avoid it.

PR #99, opened 2019-01-03
Fix push to use new push gateway api, instead of deprecated one.

### LINKED ISSUE #98: Push uses legacy API
The API that this client uses to push metrics to a push gateway was deprecated and then removed in this [commit](https://github.com/prometheus/pushgateway/commit/4c1bf91117b14b90ed17bc39b91ec49d721d716a#diff-7ddfb3e035b42cd70649cc33393fe32c). Push from this client to newer versions of the push gateway fail with a `404 Not Found` error.

cwndrws:
Fixes #98

### DIFF
--- a/lib/prometheus/client/push.rb
+++ b/lib/prometheus/client/push.rb
@@ -14,8 +14,8 @@ module Prometheus
     # Pushgateway.
     class Push
       DEFAULT_GATEWAY = 'http://localhost:9091'.freeze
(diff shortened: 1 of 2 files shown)


beorn7:
Apologies. This slipped through the radar of the maintainers. The problem has been fixed in #102 by now. Closing this. Thanks for your contribution anyway.

PR #97, opened 2018-12-20
Fix pushgateway paths.

butcher:
Hi @grobie, please review this fix, cause currently `ruby_client` is broken.

Since pushgateway released v0.7.0 deprecated `/metrics/jobs/` interface
was removed and only singularized `/metrics/job/` support now.

See [pushgateway release notes](https://github.com/prometheus/pushgateway/releases/tag/v0.7.0) and PR that [removes deprecated API](https://github.com/prometheus/pushgateway/pull/227)

PR #101, opened 2019-01-18
Correct HTTP Push end point

meetme2meat:
As of [0.7.0](https://github.com/prometheus/pushgateway/blob/d5a56ba224c85c1c027fb069a0f421ef150a9f00/CHANGELOG.md#070--2018-12-07) the HTTP push endpoint has been changed from /jobs/ ->  /job/. 

This pull request attempt to correct that.

### DIFF
--- a/lib/prometheus/client/push.rb
+++ b/lib/prometheus/client/push.rb
@@ -14,8 +14,8 @@ module Prometheus
     # Pushgateway.
     class Push
       DEFAULT_GATEWAY = 'http://localhost:9091'.freeze
-      PATH            = '/metrics/jobs/%s'.freeze
-      INSTANCE_PATH   = '/metrics/jobs/%s/instances/%s'.freeze
+      PATH            = '/metrics/job/%s'.freeze
+      INSTANCE_PATH   = '/metrics/job/%s/instances/%s'.freeze
       SUPPORTED_SCHEMES = %w(http https).freeze
 
       attr_reader :job, :instance, :gateway, :path
--- a/spec/prometheus/client/push_spec.rb
+++ b/spec/prometheus/client/push_spec.rb
@@ -61,19 +61,19 @@ describe Prometheus::Client::Push do
     it 'uses the default metrics path if no instance value given' do
       push = Prometheus::Client::Push.new('test-job')
 
-      expect(push.path).to eql('/metrics/jobs/test-job')
+      expect(push.path).to eql('/metrics/job/test-job')
     end
 
     it 'uses the full metrics path if an instance value is given' do
       push = Prometheus::Client::Push.new('bar-job', 'foo')
 
-      expect(push.path).to eql('/metrics/jobs/bar-job/instances/foo')
+      expect(push.path).to eql('/metrics/job/bar-job/instances/foo')
     end
 
     it 'escapes non-URL characters' do
       push = Prometheus::Client::Push.new('bar job', 'foo <my instance>')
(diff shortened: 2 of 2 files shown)


lib/prometheus/client/push.rb
```diff
@@ -14,8 +14,8 @@ module Client
     # Pushgateway.
     class Push
       DEFAULT_GATEWAY = 'http://localhost:9091'.freeze
-      PATH            = '/metrics/jobs/%s'.freeze
-      INSTANCE_PATH   = '/metrics/jobs/%s/instances/%s'.freeze
+      PATH            = '/metrics/job/%s'.freeze
+      INSTANCE_PATH   = '/metrics/job/%s/instances/%s'.freeze
```
beorn7: s/instances/instance/

spec/prometheus/client/push_spec.rb
```diff
@@ -61,27 +61,27 @@
...
 
-      expect(push.path).to eql('/metrics/jobs/test-job')
+      expect(push.path).to eql('/metrics/job/test-job')
     end
 
     it 'uses the full metrics path if an instance value is given' do
       push = Prometheus::Client::Push.new('bar-job', 'foo')
 
-      expect(push.path).to eql('/metrics/jobs/bar-job/instances/foo')
+      expect(push.path).to eql('/metrics/job/bar-job/instances/foo')
```
beorn7: s/instances/instance/

spec/prometheus/client/push_spec.rb
```diff
@@ -61,27 +61,27 @@
...
     end
 
     it 'escapes non-URL characters' do
       push = Prometheus::Client::Push.new('bar job', 'foo <my instance>')
 
-      expected = '/metrics/jobs/bar%20job/instances/foo%20%3Cmy%20instance%3E'
+      expected = '/metrics/job/bar%20job/instances/foo%20%3Cmy%20instance%3E'
```
beorn7: s/instances/instance/

beorn7:
Note that this is the same as https://github.com/prometheus/client_ruby/pull/99 and https://github.com/prometheus/client_ruby/pull/97 .

@grobie could you shepherd this into a solution? Note that Pushgateway v0.7 doesn't support the legacy paths anymore.

lib/prometheus/client/push.rb
meetme2meat: @beorn7 I can change that but I do seem to find the anything mentioned in the README.md regarding the `instances` thing?

Eric-Fontana-Bose:
Shouldn’t it be instance (singular) not instances ?  See https://github.com/prometheus/pushgateway/pull/227/files

beorn7:
It can be anything, but with the current endpoint, if you use `/instances/foo`, you will push with a label `instances="foo"` rather than `instance="foo"`.

Thus, if you still want to push an instance tag (which is not recommended in general), you need to use `/instance/` in the path, even if `/instances/` is technically possible.

beorn7:
The minimum fix of client_ruby is to do s/jobs/job/ and s/instances/instance/ without changing the function signature. Deprecation of pushing with an instance label can be done separately.

@grobie is the maintainer and has to shepherd this.

beorn7:
I talked to @grobie and @stuartnelson3 and created #102 with the suggested changes. Thus, this has been superseded. My apologies for the slow processing. We should have done a better job as maintainers.

PR #102, opened 2019-01-25
Update push paths

beorn7:
Legacy push paths have been removed from PGW 0.7.


beorn7:
I did this to get things going. This supersedes #101 , #99, and #97.

@grobie: @stuartnelson3 promised he would review this to make sure it works from the Ruby perspective, which is not my core competence.

grobie:
Thanks.

Issue #107, opened 2019-05-01
Consider exporting unaggregated per-pid metrics by default


Sinjo:
We had a [couple of](https://github.com/prometheus/client_ruby/pull/95#discussion_r262055780) [comments](https://github.com/prometheus/client_ruby/pull/95#issuecomment-471575689) on this point, and it’s fair. The point about “no obvious right way to aggregate gauges in all cases” is a solid argument.

We need to decide a stance on this both for gauges specifically and for metrics in general (i.e. should gauges be a special case where we don't aggregate by default).

This issue should be renamed and updated once we decide that stance.

Sinjo:
Right, I've been digging through all the relevant code and doing some thinking on which approach to take.

Looking at the aggregations [supported by the Python client](https://github.com/prometheus/client_python/tree/67402133513b3658b1b1af658066403c2f2f8cdd#multiprocess-mode-gunicorn) (scroll down to 'Three: Instrumentation'), I think we should start by supporting the 'all' mode. The 'liveall' and 'livesum' modes sound interesting and potentially useful, but I'm confident we can add them later with no breaking changes (i.e. I'm not making them part of the path to 1.0).

My plan is to:
  - Add `:all` as a new aggregation mode in `DirectFileStore`. When specified, it will add a `"pid"` label to the metric and populate it with the process ID that exported that value.
  - Make `:all` the default aggregation mode for gauges only.
  - Make `"pid"` a [reserved label for all metric types](https://github.com/prometheus/client_ruby/blob/6ea675ef633c40a45643dedd76eae0b87c32dbcf/lib/prometheus/client/metric.rb#L65). While it will only be the default for gauges, my planned approach and current implementation (#127) support `:all` for any type of metric, and I don't want to overwrite people's label values silently. This will mean changing the subclasses of `Prometheus::Client::Metric` to append their reserved labels to those from the base class, rather than [overriding them completely](https://github.com/prometheus/client_ruby/blob/6ea675ef633c40a45643dedd76eae0b87c32dbcf/lib/prometheus/client/histogram.rb#L112).

I'm happy to discuss any of those choices and change tack if there's a good reason to.

lawrencejones:
Fwiw, providing a pid label has been the only legitimate way I've found to use gauges with the multi-process library. Reckon it's the only semantically valid approach!

SuperQ:
We use a `worker` label and assign the unicorn worker ID to our per-process metrics. This avoids some metric name churn on worker restarts.

[...]

PR #125, opened 2019-05-23
Add upgrade documentation for 0.9 to 0.10

rsetia:
@Sinjo 

Documentation describes differences between 0.9 and 1.x which will have breaking changes due to new multiprocess code.


Sinjo:
Hi @rsetia,

Thanks for sending this over! I've had a quick skim and it looks good overall. There are a couple of adjustments I'd like to make (e.g. I think it was either Ruby 2.1 or 2.2 we landed on), which I'll get to the week after next. Next week I'm not going to be working on Prometheus I'm afraid, and today I'm focused on #107.

Thanks again for your contribution!

Sinjo:
@dmagliola I've made a few tweaks and added some sections based on further work we've done. You good to give this a review?

dmagliola:
Merging even though CI is red. That's unrelated to these changes, and it's fixed by #152 which is already merged, so `master` should remain green

Issue #111, opened 2019-05-01
Remove flexible label support from Rack middleware


Sinjo:
Supporting flexible labels in our out-the-box Rack middleware commits us to maintaining what is currently quite a [confusing API](https://github.com/prometheus/client_ruby/pull/95#discussion_r233501944).

@dmagliola discussed a few ways to make the API less weird, but they always resulted in the middleware accepting multiple lambdas for custom behaviour as arguments and having almost no behaviour provided out the box - sort of defeating the purpose!

We can always come back and add a better implementation of this functionality later, but it will be a pain to take this version of it away.

I’m in favour of only supporting our fixed set of labels in the Rack middleware we provide, and having a README section advising people to do their own thing if they want something more sophisticated.

PR #126, opened 2019-05-27
Update example README for #111

swalberg:
I was trying to modify the path that's exported and didn't realize it was _just_ changed. The documentation in the example project still reflects the old way.

Gave a short example of how I did it -- I realize subclassing a protected method is dangerous but there are no other hooks.

FWIW my use case was that the default regular expressions did not cover my url structure and I needed to add another one. I did not need to add more labels.

### DIFF
--- a/examples/rack/README.md
+++ b/examples/rack/README.md
@@ -44,21 +44,20 @@ Prometheus server can be used to [play around with the metrics][rate-query].
 The example shown in [`config.ru`](config.ru) is a trivial rack application
 using the default collector and exporter middlewares.
 
-In order to use custom label builders in the collector, change the line to
-something like this:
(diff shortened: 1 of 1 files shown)


Sinjo:
Hi @swalberg,

Thanks for taking time to write this and send it over!

I'm going to merge as is and make a couple of tweaks to caveat that extending the class isn't something we support as an API, but that it is an option.

To give a bit of extra clarity on the Rack middleware situation: making it work with the the up-front label declaration left us with an API we thought was more confusing than it was worth.

I am keen to come back to it into the future, but for now I'm focusing on the road to 1.0, which means focusing on more breaking changes like #107 (which I'm working on today).

PR #127, opened 2019-05-31
Support :all as an aggregation mode in DirectFileStore

Sinjo:
This is the beginning of the work for #107. I'm going to detail the plan I came up with in there.

Not mega happy with this implementation, so I've marked this PR as a draft for now. Will see if I can come up with anything better.

### DIFF
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -26,7 +26,7 @@ module Prometheus
 
       class DirectFileStore
         class InvalidStoreSettingsError < StandardError; end
-        AGGREGATION_MODES = [MAX = :max, MIN = :min, SUM = :sum]
+        AGGREGATION_MODES = [MAX = :max, MIN = :min, SUM = :sum, ALL = :all]
         DEFAULT_METRIC_SETTINGS = { aggregation: SUM }
 
         def initialize(dir:)
@@ -140,6 +140,10 @@ module Prometheus
           end
 
           def store_key(labels)
+            if @values_aggregation_mode == ALL
+              labels[:pid] = process_id
+            end
+
             labels.map{|k,v| "#{CGI::escape(k.to_s)}=#{CGI::escape(v.to_s)}"}.join('&')
           end
 
@@ -168,6 +172,8 @@ module Prometheus
               values.max
             elsif @values_aggregation_mode == MIN
               values.min
+            elsif @values_aggregation_mode == ALL
+              values.first
             else
               raise InvalidStoreSettingsError,
                     "Invalid Aggregation Mode: #{ @values_aggregation_mode }"
--- a/README.md
+++ b/README.md
@@ -256,6 +256,15 @@ class MyComponent
 end
 ```
 
+### Reserved labels
+
+The following labels are reserved by the client library, and attempting to use them in a
+metric definition will result in a
+`Prometheus::Client::LabelSetValidator::ReservedLabelError` being raised:
+
+  - `:job`
+  - `:instance`
+  - `:pid`
 
 ## Data Stores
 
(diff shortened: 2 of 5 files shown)


lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -120,16 +120,30 @@ def all_values
                     [k.to_sym, vs.first]
                   end.to_h
 
+                  if @values_aggregation_mode == ALL
+                    #  TODO: Can I do this in a less gross way?
+                    pid = /(\d+)\.bin$/.match(file_path)[1]
```
dmagliola: I'd add the label when writing to the file, and then you get it out for free when reading here, no need to do this. 
Basically, add it in `store_key` below

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -120,16 +120,30 @@ def all_values
...
+              stores_data.each do |label_set, values|
+                stores_data[label_set] = values.first
+              end
+
+              stores_data
+            else
+              # Aggregate all the different values for each label_set
+              stores_data.each_with_object({}) do |(label_set, values), acc|
+                acc[label_set] = aggregate_values(values)
+              end
```
dmagliola: I'd also do this in the `aggregate_values` method below. 

```
           elsif @values_aggregation_mode == ALL
              values.min	              values.first
```

Sinjo: I have no idea what I was thinking when I wrote my version. This is so obviously the right way to do it.

Updated!

lib/prometheus/client/data_stores/direct_file_store.rb
Sinjo: Yeah on reflection that seems better. I wasn't sure how I felt about writing out a label that we were injecting, but I can't see any serious downside to it.

Sinjo:
Before I take this out of draft, I'm going to make `pid` into a globally reserved label as I mentioned in #107.

I think it's cleaner to disallow it completely rather than saying "If you're using the gem in this particular mode then it's okay otherwise we'll overwrite it, sucks to be you!"

dmagliola:
Definitely the right call to make it reserved.

Sinjo:
I've marked this as ready for review. As far as I can tell, the build failures are a (hopefully transient) RVM issue in the JRuby build.

```
$ rvm get stable
Downloading https://get.rvm.io
Downloading https://raw.githubusercontent.com/rvm/rvm/master/binscripts/rvm-installer.asc
Verifying /home/travis/.rvm/archives/rvm-installer.asc
gpg: Signature made Tue 23 Apr 2019 04:50:26 PM UTC using RSA key ID 39499BDB
gpg: Good signature from "Piotr Kuczynski <piotr.kuczynski@gmail.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 7D2B AF1C F37B 13E2 069D  6956 105B D0E7 3949 9BDB
GPG verified '/home/travis/.rvm/archives/rvm-installer'
Downloading https://github.com/rvm/rvm/archive/.tar.gz
curl: (22) The requested URL returned error: 404 Not Found
Could not download 'https://github.com/rvm/rvm/archive/.tar.gz'.
  curl returned status '22'.
Downloading https://bitbucket.org/mpapis/rvm/get/.tar.gz
curl: (22) The requested URL returned error: 404 Not Found
Could not download 'https://bitbucket.org/mpapis/rvm/get/.tar.gz'.
  curl returned status '22'.
Could not update RVM, please report to https://github.com/rvm/rvm/issues
The command "rvm get stable" failed and exited with 22 during .
Your build has been stopped.
```

Will take a look later if it turns out not to be transient.

PR #130, opened 2019-06-14
Update README

ganmacs:
Summary#get returns Hash instance

https://github.com/prometheus/client_ruby/blob/3652cf70c9f026d59d64594aa5fa03e4a740128a/lib/prometheus/client/summary.rb#L31-L33

Signed-off-by: Yuta Iwama <ganmacs@gmail.com>


dmagliola:
Confirmed, that's the correct code.
Sorry for the delay!

PR #131, opened 2019-06-14
Export per-pid values from gauges by default

### LINKED ISSUE #107: Consider exporting unaggregated per-pid metrics by default

Sinjo:
This changes `DirectFileStore` to use the `:all` aggregation mode for gauges by default.

I'm considering adding a config option to the store to allow the default to be overridden, for use in cases where people want to provide their own label that distinguishes between processes but changes less often than process IDs do.

I'll try to add a commit early next week for that.

Fixes #107

### DIFF
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -28,6 +28,7 @@ module Prometheus
         class InvalidStoreSettingsError < StandardError; end
         AGGREGATION_MODES = [MAX = :max, MIN = :min, SUM = :sum, ALL = :all]
         DEFAULT_METRIC_SETTINGS = { aggregation: SUM }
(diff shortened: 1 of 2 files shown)


Sinjo:
Following a conversation with @dmagliola, I've decided to leave that extra commit off.

It's a nice-to-have, it'd be a non-breaking change if we added it in the future, and the current approach isn't particularly onerous.

Taking this out of draft.

PR #138, opened 2019-06-18
Fixes bug that prevented store settings from being passed to the metrics

jbernardo95:
This PR fixes a bug in the registry that prevented the user from configuring the store settings of new metrics.

cc @dmagliola @Sinjo


dmagliola:
🤦‍♂ My bad!
Good find, thank you!!

PR #142, opened 2019-07-04
rubocop

lawrencejones:
Re-apply rubocop to our codebase. This is broken into many auto-correct commits, plus hand-edited commits.

No functionality has been changed.


dmagliola:
I'm going to close this as stale, the code has changed too much since it was attempted.

After a quick chat with @Sinjo, though, I feel that "some very small subset of Rubocop for very narrow styling things only" would be somewhat positive, but in a codebase this size, with these few maintainers, where I don't think we've ever had a single argument in a PR about code styling... Rubocop probably adds more overhead than value, IMO.

Maybe when rubyfmt comes around we should give that a try. Feels a lot narrower than "Rubocop default opinions"

PR #144, opened 2019-07-17
Avoid revalidating labelset keys on each metric observation

dmagliola:
When declaring a metric, we declare what label keys will be used for it.
At that point, we validate that those are all valid keys (symbols,
match a given regex, etc).

Then, on each observation of the metric, we validate that the keys passed
in for the labels match the ones we were originally expecting, to make
sure all of those labels were set, but no others.

We were also, at that point, validating that the keys passed in are valid.
This validation is pretty slow, and it's redundant, since keys that aren't
valid won't match the expected ones anyway, so we can just compare just
that those match. This has quite a pronounced effect on performance.

### DIFF
--- a/lib/prometheus/client/label_set_validator.rb
+++ b/lib/prometheus/client/label_set_validator.rb
@@ -36,15 +36,20 @@ module Prometheus
       def validate_labelset!(labelset)
         return labelset if @validated.key?(labelset.hash)
 
-        validate_symbols!(labelset)
-
-        unless keys_match?(labelset)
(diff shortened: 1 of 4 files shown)


dmagliola:
Turns out, this has a massive impact if we remove the cache (which is one of the options we were contemplating to solve one of the outsanding issues). Keeping the cache and doing this means we don't gain that much, so it doesn't make sense to have this isolated PR. Closing, and this code will show up in another PR, if that's the direction we go.

PR #147, opened 2019-07-24
Test that stores return floats and fix DirectFileStore to do so

Sinjo:
Up to now, we've had some subtle behavioural differences between stores;
one of which was that stores weren't consistent about what they returned
for label sets that hadn't been initialised.

This commit adds some shared examples that enforce consistent behaviour
between the stores and fixes DirectFileStore to follow that behaviour.


Sinjo:
> Finally, this is the last inconsistency between stores, right?
I sure hope so!

PR #148, opened 2019-07-26
Reopen files in DirectFileStore when the process has forked

### LINKED ISSUE #119: Detect PID changes and open new store in `DirectFileStore`
Since we're creating a store that's designed for use from multiple processes, it would be nice to more gracefully handle PID changes caused by forking gracefully.

Currently, if you wrote to a metric in the original process, the new process will continue writing to the metric file opened by the original process, as the file handle [is kept around](https://github.com/gocardless/prometheus_client_ruby/blob/5dce3e5855f2aa67c65874c953878e124c23edcd/lib/prometheus/client/data_stores/direct_file_store.rb#L147).

We'd need some sort of cache-busting-esque mechanism that caused a new metrics file to be opened when a PID change was detected.

**Note that you shouldn't be writing any metrics before you fork! Implementing this would be rounding off a sharp edge.**

Sinjo:
Currently, if you fork after altering a metric value when using
DirectFileStore, the child process will have a handle to the original
process's metric store.

This commit detects checks for that condition on any instrumentation
event and opens a new metric file for the new PID.

Fixes #119

### DIFF
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -154,7 +154,12 @@ module Prometheus
           end
 
           def internal_store
-            @internal_store ||= FileMappedDict.new(filemap_filename)
+            if @store_opened_by_pid != process_id
+              @store_opened_by_pid = process_id
+              @internal_store = FileMappedDict.new(filemap_filename)
+            else
+              @internal_store
+            end
           end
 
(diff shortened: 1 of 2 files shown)


dmagliola:
Note that an empty block in `ips` gives you about  13M operations (rather than 11M), so I wouldn't even trust that 0.1µs, it's probably even less noticeable than that

PR #153, opened 2019-09-02
Fixing the Registry threading/design issues 

### LINKED ISSUE #151: No response body from `/metrics` with multiple servers
# Code first
```RUBY
require "prometheus/middleware/collector"
require "prometheus/middleware/exporter"

module Sinatra
  class Api < Base   
    # Used for Prometheus
    use Rack::Deflater
    use Prometheus::Middleware::Collector
    use Prometheus::Middleware::Exporter

    configure do
      set :raise_errors, true
    end
  end
end
```

# Technology
We use `JRuby-9.2.4.1` and [Trinidad server](https://github.com/trinidad/trinidad)

## Setup
We have 2 servers running the  application with the same ruby version along with same Trinidad server. On front of them HAProxy as a load balancer.

# The issue
When you call the `/metrics` page, HAProxy forward the request to the available server, if both are available then selection is a bit random.
The issue is, on one server you get the response and you can see the metrics with OpenMetrics format, on the other, you get `200` status code back but no body provided.
Just blank page.

`curl`ing the servers, one is working pretty fine and response body has the metrics, other works pretty fine response body is empty.
Logs are also fine.
Didn't really go deep in the application to see what's wrong (maybe threading issue with the datastores)
But wanted to ask you guys first, maybe you had this issue before or have some hints

ahmgeek:
# What does it fix #151 
There are two issues with the current implementation:
1. In the collector middleware
2. A design issue with the registry

I will go over them one by one

## The collector
https://github.com/prometheus/client_ruby/blob/80adc307f90d59f977324c3e5a38bcc9f980cb5f/lib/prometheus/middleware/collector.rb#L36-L37

These `init_*` methods get called every single time a request happens
https://github.com/prometheus/client_ruby/blob/b3bbe23ae8b9bdf3379699e310a8450a46efa1fe/lib/prometheus/middleware/collector.rb#L46-L47
As you can see, every request we try to register the default `http_*` metrics.
If you think this is not an issue with MRI, it is actually, it raises `AlreadyRegisteredError` with MRI
This is the spec https://github.com/prometheus/client_ruby/pull/153/commits/d4f4e3226cec8d6c42a704b222457c9702fccd83 that produces the bug.

### The fix
Check before adding the metric to the registry, if the metric is already there, use it, otherwise push the metric to the registry.
Here https://github.com/prometheus/client_ruby/pull/153/commits/fb4504e5aae5a22caf9fb2e8de129651ad641406 I introduce the `registry.get(metric)`

I think Daniel mentioned something regards symbols/strings. I would vote for strings, I can try and fix this too but it will be a breaking change if I understand correctly? maybe we can have a coercing method under the hood and warn about the deprecation or something.

## The registry
Turned out, in a multi-threaded world, two threads could race to register the metric, and bypass the `@registry.get(metric)` check. So one would register the metric and the other would raise `AlreadyRegisteredError`.
The way I see it if multiple threads try to register the same metric (with locking introduced), if the metric is registered, there should be no errors raised, just return the metric and use it to perform the preferred action.
Hence here https://github.com/prometheus/client_ruby/pull/153/commits/80adc307f90d59f977324c3e5a38bcc9f980cb5f I changed from rubt's mutex, to something more efficient for the use case.

The specs for thread safety was relying on the `raise` to make it thread-safe, for me, this design is more of a workaround. If the registry is thread-safe, it means only one metric will be written to the registry at all times.

## The registry again
I re-designed in the last commit the registry to be used as `singleton` object. This way we don't need the rather ~bad~ not thread safe memoization.

## Points to discuss
@dmagliola Please let me know how is the new design works, maybe I understand things wrongly, or my implementation has flaws. Also if you can put me in the loop while you discuss this with your team that would be great.

### DIFF
--- a/lib/prometheus/client.rb
+++ b/lib/prometheus/client.rb
@@ -8,7 +8,7 @@ module Prometheus
   module Client
     # Returns a default registry object
     def self.registry
-      @registry ||= Registry.new
+      Registry.instance
     end
 
     def self.config
--- a/lib/prometheus/client/metric.rb
+++ b/lib/prometheus/client/metric.rb
@@ -1,6 +1,7 @@
 # encoding: UTF-8
 
 require 'thread'
+require 'prometheus/client'
 require 'prometheus/client/label_set_validator'
 
 module Prometheus
--- a/lib/prometheus/client/registry.rb
+++ b/lib/prometheus/client/registry.rb
@@ -5,12 +5,15 @@ require 'prometheus/client/summary'
 require 'prometheus/client/gauge'
 require 'prometheus/client/histogram'
 require 'concurrent'
+require 'singleton'
 
 module Prometheus
   module Client
     # Registry
     class Registry
       class AlreadyRegisteredError < StandardError; end
+
+      include Singleton
 
       def initialize
         @metrics = {}
--- a/lib/prometheus/client/registry.rb
+++ b/lib/prometheus/client/registry.rb
@@ -1,11 +1,10 @@
 # encoding: UTF-8
 
-require 'thread'
-
 require 'prometheus/client/counter'
 require 'prometheus/client/summary'
 require 'prometheus/client/gauge'
 require 'prometheus/client/histogram'
+require 'concurrent'
 
 module Prometheus
   module Client
@@ -15,16 +14,15 @@ module Prometheus
 
       def initialize
         @metrics = {}
-        @mutex = Mutex.new
+        @lock    = Concurrent::ReentrantReadWriteLock.new
       end
 
       def register(metric)
         name = metric.name
 
-        @mutex.synchronize do
-          if exist?(name.to_sym)
-            raise AlreadyRegisteredError, "#{name} has already been registered"
-          end
+        @lock.with_write_lock do
+          return metric if exist?(name.to_sym)
+
           @metrics[name.to_sym] = metric
         end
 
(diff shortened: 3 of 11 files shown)


ahmgeek:
@dmagliola please take a look and merge it if you think it's fine this fixes the issue within the collector.

dmagliola:
Hi @ahmgeek, thank you for your PR!
I'm currently on vacation. I tried to look into this, but it's quite involved, and some of these changes require discussions with other members of the team (like returning a metric instead of erroring if registering it twice).

This PR also highlights a few other small problems, like the confusion between strings and symbols in the `Registry`, which we should also fix. 
I'll be looking into this near the end of September.

Sorry for the delay, and thanks for the good work!

ahmgeek:
@dmagliola Oh man, am really sorry to disturb you during the vacation, please forget about it. Have fun and enjoy the vacation.
We can discuss it later. 😘

ahmgeek:
Unfortunately, on local development machine randomly the `/metrics` endpoint returns the initialized metrics in a freezed way i.e never increments the counters for instance.

In production its always freezed, there's no way to increment them.
I will close this PR now, I think we will go for an internal implementation within my company (targeting jRuby mainly)

Thanks ❤️

ahmgeek:
@mos3abof  ميصحش كده خالص يا كابتن :)

Issue #141, opened 2019-06-27
RFC: Counter initialisation


jbernardo95:
# Problem

The problem this proposal tries to solve is the one described [here](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics).

In essence it can be summarised as:

> Time series that are not present until something happens are difficult to deal with, as the usual simple operations are no longer sufficient to correctly handle them. 

An example of this is when we try to calculate the `increase()` of a counter from the beggining. If the first counter value is not 0 the result of calculating increase is going to be wrong: `increase([1, 1, 2, 2]) # returns 1 but should return 2 given that the counter went from 0 to 2`.

[This](https://github.com/prometheus/prometheus/issues/1853) and [this](https://github.com/prometheus/prometheus/issues/1673) are related issues.

Note that this problem does not apply to Gauges and as such I ignore them in the rest of the RFC.

# Solution

The solution to this problem is described in the Prometheus website:

> export 0 (or NaN, if 0 would be misleading) for any time series you know may exist in advance.

Note that this solution does not cover the cases where the Prometheus scrape window did not catch the counter value at 0.

For metrics without labels this is as simple as initialising the counter value to 0 and exposing it from the beggining. In fact, this is something that [other client libraries are already doing](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics).

On the other hand for metrics with several labels the only way to initialise counters for all these labels is to know the label values in advance, so that we can initialise all possible counters/label combinations. For example, if we have a counter with two labels, `app` and `env`. And the possible values for each of these labels are:

- `app`
  - `frontend`
  - `backend`
- `env`
  - `production`
  - `staging`

Then we would have to initialise the following counters to 0:

- `app=frontend,env=production`
- `app=frontend,env=staging`
- `app=backend,env=production`
- `app=backend,env=staging`

# Implementation

My proposal to solve this problem is to add a mechanism to the client that allows the user to list all possible label values upfront so that the client can initialise all the counters to 0.

Ways this could be implemented:

```ruby
# Adding yet another parameter to the metric
Counter.new(
  :http_requests_total, 
  docstring: '...', 
  labels: [:app, :env],
  label_values: {
    app: [:frontend, :backend],
    env: [:production, :staging]
  }
)

# Creating a method to provide the label values an initialise the counters
counter =
  Counter.new(
    :http_requests_total, 
    docstring: '...', 
    labels: [:app, :env]
  )

counter.init({
[...]

PR #156, opened 2019-10-01
Implements #init_label_set method for all metric classes

jbernardo95:
This a draft PR of a solution for the problem described in #141.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,12 @@ class MyComponent
 end
 ```
 
+### `init_label_set`
+
+The time series of a metric are not initialized until something happens. For counters, for example, this means that the time series do not exist until the counter is incremented for the first time.
+
+To get around this problem the client provides the `init_label_set` method that can be used to initialise the time series of a metric for a given label set.
+
 ### Reserved labels
 
 The following labels are reserved by the client library, and attempting to use them in a
--- a/lib/prometheus/client/histogram.rb
+++ b/lib/prometheus/client/histogram.rb
@@ -94,6 +94,16 @@ module Prometheus
         end
       end
 
+      def init_label_set(labels)
+        base_label_set = label_set_for(labels)
+
+        @store.synchronize do
+          (buckets + ["+Inf", "sum"]).each do |bucket|
+            @store.set(labels: base_label_set.merge(le: bucket.to_s), val: 0)
+          end
+        end
+      end
+
       private
 
       # Modifies the passed in parameter
--- a/lib/prometheus/client/metric.rb
+++ b/lib/prometheus/client/metric.rb
@@ -55,6 +55,10 @@ module Prometheus
                        store_settings: @store_settings)
       end
 
+      def init_label_set(labels)
+        @store.set(labels: label_set_for(labels), val: 0)
+      end
+
       # Returns all label sets with their values
       def values
         @store.all_values
--- a/lib/prometheus/client/summary.rb
+++ b/lib/prometheus/client/summary.rb
@@ -45,6 +45,15 @@ module Prometheus
         end
       end
 
+      def init_label_set(labels)
+        base_label_set = label_set_for(labels)
+
+        @store.synchronize do
+          @store.set(labels: base_label_set.merge(quantile: "count"), val: 0)
+          @store.set(labels: base_label_set.merge(quantile: "sum"), val: 0)
+        end
+      end
+
       private
 
       def reserved_labels
--- a/spec/prometheus/client/counter_spec.rb
+++ b/spec/prometheus/client/counter_spec.rb
@@ -79,4 +79,16 @@ describe Prometheus::Client::Counter do
       end.to change { counter.get }.by(100.0)
     end
   end
+
+  describe '#init_label_set' do
+    let(:expected_labels) { [:test] }
+
+    it 'initializes the metric for a given label set' do
+      expect(counter.values).to eql({})
+
+      counter.init_label_set(test: 'value')
+
+      expect(counter.values).to eql({test: 'value'} => 0.0)
+    end
+  end
 end
--- a/spec/prometheus/client/gauge_spec.rb
+++ b/spec/prometheus/client/gauge_spec.rb
@@ -161,4 +161,16 @@ describe Prometheus::Client::Gauge do
       end.to change { gauge.get }.by(-100.0)
     end
   end
+
+  describe '#init_label_set' do
+    let(:expected_labels) { [:test] }
+
+    it 'initializes the metric for a given label set' do
+      expect(gauge.values).to eql({})
+
+      gauge.init_label_set(test: 'value')
+
+      expect(gauge.values).to eql({test: 'value'} => 0.0)
+    end
+  end
 end
--- a/spec/prometheus/client/histogram_spec.rb
+++ b/spec/prometheus/client/histogram_spec.rb
@@ -118,4 +118,20 @@ describe Prometheus::Client::Histogram do
       )
     end
   end
+
+  describe '#init_label_set' do
+    let(:expected_labels) { [:status] }
+
+    it 'initializes the metric for a given label set' do
+      expect(histogram.values).to eql({})
+
+      histogram.init_label_set(status: 'bar')
+      histogram.init_label_set(status: 'foo')
+
+      expect(histogram.values).to eql(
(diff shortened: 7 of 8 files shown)


lib/prometheus/client/histogram.rb
```diff
@@ -94,6 +94,23 @@ def values
...
+            bucket_label_set = base_label_set.dup
+            bucket_label_set[:le] = bucket.to_s
+            sum_label_set = base_label_set.dup
+            sum_label_set[:le] = "sum"
+
+            @store.set(labels: bucket_label_set, val: 0)
+            @store.set(labels: sum_label_set, val: 0)
+          end
+        end
+      end
```
jbernardo95: I just realised that this client does not provide a `_count` metric as part of the histogram. Is this true? Or did I miss it?

lib/prometheus/client/metric.rb
```diff
@@ -55,6 +55,10 @@ def with_labels(labels)
                        store_settings: @store_settings)
       end
 
+      def init_label_set(labels)
+        @store.set(labels: label_set_for(labels), val: 0)
+      end
```
jbernardo95: I'm not a fan of having a default implementation and some sub classes overriding it, but this seems to be on par with the `#with_labels` implementation.

lib/prometheus/client/histogram.rb
SuperQ: Hrmm, good question, there should be a count, it's required for histograms. Maybe it's auto-generated from the `+Inf` bucket (they should be the same value).

dmagliola: @jbernardo95 The client does provide a count in the export to prometheus, it just doesn't need to explicitly store it in the hash, because as SuperQ said, it's the same as the +Inf value.

This is where the magic happens: https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/client/formats/text.rb#L67

lib/prometheus/client/metric.rb
dmagliola: Yeah, sadly, Summaries and Histograms are doing "magic" to the labels to be able to basically wedge more complex objects into a plain hash we can serialise to disk. 
It's a bit of a hack, admittedly, but I couldn't think of anything better...

lib/prometheus/client/histogram.rb
```diff
@@ -94,6 +94,23 @@ def values
...
+        @store.synchronize do
+          [*@buckets, "+Inf"].each do |bucket|
+            # This is basically faster than doing `.merge`
+            bucket_label_set = base_label_set.dup
+            bucket_label_set[:le] = bucket.to_s
+            sum_label_set = base_label_set.dup
+            sum_label_set[:le] = "sum"
+
+            @store.set(labels: bucket_label_set, val: 0)
+            @store.set(labels: sum_label_set, val: 0)
```
dmagliola: I found this code a bit hard to read, because you're setting `sum` for each of the buckets in the array.
I think it'd be better if we did `[*@buckets, "+Inf", "sum"]`, instead of `[*@buckets, "+Inf"]`. I may be missing something, but that seems like it'd do it?

Also, i'd use `base_label_set.merge(le: bucket)` instead of the "dup" + setting the key.
It's way more readable, and here we don't care about the tiny performance difference like we do when observing.

dmagliola: Not sure is that comment made much sense, sorry :)
I'm proposing this:

```
@store.synchronize do
  (buckets + ["+Inf", "sum"]).each do |bucket|
    @store.set(labels: base_label_set.merge(le: bucket.to_s), val: 0)
  end
end
```

Would that work? Does the test still pass with the same result?

jbernardo95: > you're setting sum for each of the buckets in the array.

Yeah, there is no need for this! And I agree your suggestions is a lot more readable.

lib/prometheus/client/histogram.rb
jbernardo95: Makes sense 👍

jbernardo95:
@dmagliola Feedback addressed let me know what you think!

README.md
```diff
@@ -256,6 +256,49 @@ class MyComponent
 end
 ```
 
+### `init_label_set`
+
+The time series of a metric are not initialiazed until something happens. For counters, for example, this means that the time series do not exist until the counter is increamented for the first time.
```
dmagliola: /s/initialiazed/initialized/

dmagliola: /s/increamented/incremented/

README.md
```diff
@@ -256,6 +256,49 @@ class MyComponent
...
+counter =
+  Counter.new(
+    :http_requests_total,
+    docstring: '...',
+    labels: [:app, :env, :version],
+  )
+
+# After declaring the counter nothing is exported because the time series have not been initialized
+
+# We call increment for the first time and the time serie for the label set provided is initialiazed
```
dmagliola: /s/initialiazed/initialized/

README.md
```diff
@@ -256,6 +256,49 @@ class MyComponent
...
+counter.increment(labels: {app: "frontend", env: "production", version: "v1"})
+
+# Client starts exporting
+# http_requests_total{app="frontend", env="production", version="v1"} 1
+
+# We call increment again
+counter.increment(labels: {app: "frontend", env: "production", version: "v1"})
+
+# Client exports
+# http_requests_total{app="frontend", env="production", version="v1"} 2
```
dmagliola: I would make this shorter and with no code. I think it's easier to read if you explain the idea that you initialize a counter, it doesn't get exported until it's incremented at least once. The key sentence to explain the problem is the one below. (And i'd make it "not code") :)

README.md
```diff
@@ -256,6 +256,49 @@ class MyComponent
...
+
+# In the server when we calculate the increase of the counter, increase([1, 2]), we get 1 as a result although we increased the counter twice
+```
+
+To get around this problem the client provides the `init_label_set` method that can be used to initialises the time serie of a metric for a given label set. In the example above we would use it like so:
```
dmagliola: /s/initialises/initialise/

README.md
```diff
@@ -256,6 +256,49 @@ class MyComponent
...
+
+```ruby
+# After declaring the counter we can initiliaze the metric by calling the #init_label_set method
```
dmagliola: /s/initiliaze/initialize/

lib/prometheus/client/histogram.rb
```diff
@@ -94,6 +94,16 @@ def values
         end
       end
 
+      def init_label_set(labels)
+        base_label_set = label_set_for(labels)
+
+        @store.synchronize do
+          [*@buckets, "+Inf", "sum"].each do |bucket|
```
dmagliola: Can we do `(@buckets + ["+Inf", "sum"])`?
I find this splat harder to read

jbernardo95: Yeah sure!

README.md
jbernardo95: Yeah, the code might be a little bit verbose. I'll remove it.

jbernardo95:
@dmagliola Ready for a second round 👀

README.md
```diff
@@ -256,6 +256,12 @@ class MyComponent
 end
 ```
 
+### `init_label_set`
+
+The time series of a metric are not initialized until something happens. For counters, for example, this means that the time series do not exist until the counter is incremented for the first time.
+
+To get around this problem the client provides the `init_label_set` method that can be used to initialise the time serie of a metric for a given label set.
```
dmagliola: /s/serie/series/

(series is one of those weird words that is both singular and plural, like fish. English...)

lib/prometheus/client/histogram.rb
```diff
@@ -94,6 +94,16 @@ def values
         end
       end
 
+      def init_label_set(labels)
+        base_label_set = label_set_for(labels)
+
+        @store.synchronize do
+          (@buckets + ["+Inf", "sum"]).each do |bucket|
```
dmagliola: Damn, sorry, this is my bad. I just realized we have an `attr_reader :buckets`
Could you change this `@buckets` to just `buckets`? :D

jbernardo95: It's all about consistency 😛

README.md
jbernardo95: #til

jbernardo95:
Done 🚀

dmagliola:
Thank you @jbernardo95 for this!

Issue #143, opened 2019-07-10
`DirectFileStore` creates too many files for long running applications


maximkosov:
We use passenger to run our rails application. Worker processes recreated after every 100 requests. Recently we tried client_ruby with `DirectFileStore`. Each worker process has its own pid so depending on server load there will be tens of thousands of files in prometheus work dir after couple of hours/days when app wasn't restarted.

With 50k files in prometheus work dir /metrics route starts to be very slow with processing time about 10 seconds which could lead to prometheus scraper timeouts.

Is there any workaround for long running processes with `DirectFileStore`? One possible workaround I can think of is just restart app once in a few hours. Instead of restarting the whole application we can just wipe prometheus work directory once in a few hours, but this is looks a little bit hacky for me.

dmagliola:
This is related to issue #106, namely, how do you know whether a file is being written to by a process that's still alive. It's not exactly the same problem, since in this case, we can't get rid of files from old processes (or else counters would go down), but I have a feeling both problems will be solved in a related way.

lawrencejones:
I would think an appropriate fix for this would be to have each process open a shared lock on the file it manages, for the duration of it needing the file.

Then, every time a file store tries creating a new file, you examine all those that exist in the directory that you can exclusively lock and compact them into a new file, while destroying all those that existed before. 

Provided you think about how to do concurrency control, this would be a decent mechanism for solving it. But at this point I reckon you should probably be using an mmap system.

brian-brazil:
Is there a locking system that will work across all OSes?

lawrencejones:
> Is there a locking system that will work across all OSes?

That would certainly be a consideration, and you'd probably end up with the compaction as an optional configuration on the file store. But honestly, I'd rather see a more performant mmap approach before doing this. Just wanted to sketch out a possible implementation.

brian-brazil:
We haven't solved this, and we're using mmap over in Python. The reads were actually changed to not use mmap recently for performance.

dmagliola:
Another potential option would be to have one file per process, rather than one file per process/metric.

This doesn't fundamentally solve the issue, but it makes it orders of magnitude less problematic. The downside to this is that now each metric increment has essentially a mutex around it, so it is not the best performance in multi-threaded scenarios. 

But if, like most Ruby apps, each process is running single-threaded, or if it's not incrementing counters that often, the performance penalty will probably be negligible. Each store save should be in the order of single-digit microseconds after all.

[...]

PR #161, opened 2019-10-14
One file per process instead of per metric/process for DirectFileStore [DOESNT WORK]

dmagliola:
# This doesn't work!!

I'm just leaving it here as documentation in case someone tries to do this to solve #143
Read the commit comments for more.

NOTE: I still think this is the right thing to do (one file per process, rather than per metric/process)
It'll have more contention in multi-threaded scenarios, but it'll drastically reduce the number of files

NOTE2: We also need to add a Mutex around that shared Dict, if we're having the shared Dict! The `@lock` in `in_process_sync` needs to be shared by all `MetricStore`s

Anyway, i'm back to thinking we should either make a second `DirectFileStore` that does this (and have two separate files, instead of a param), or considerably refactor the DirectFileStore. My quick stab definitely doesn't work

### DIFF
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -67,6 +67,11 @@ module Prometheus
         class MetricStore
           attr_reader :metric_name, :store_settings
 
+          class << self
+            attr_accessor :shared_store_opened_by_pid
+            attr_accessor :shared_internal_store
+          end
+
           def initialize(metric_name:, store_settings:, metric_settings:)
             @metric_name = metric_name
             @store_settings = store_settings
@@ -164,11 +169,28 @@ module Prometheus
           end
 
           def internal_store
+            if @store_settings[:separate_files_per_metric]
+              individual_metric_internal_store
+            else
+              all_metrics_shared_internal_store
+            end
+          end
+
+          def individual_metric_internal_store
             if @store_opened_by_pid != process_id
               @store_opened_by_pid = process_id
               @internal_store = FileMappedDict.new(filemap_filename)
             else
               @internal_store
+            end
+          end
+
+          def all_metrics_shared_internal_store
+            if self.class.shared_store_opened_by_pid != process_id
+              self.class.shared_store_opened_by_pid = process_id
+              self.class.shared_internal_store = FileMappedDict.new(filemap_filename)
+            else
+              self.class.shared_internal_store
             end
           end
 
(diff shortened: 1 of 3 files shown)


saviogl:
@dmagliola Do you know if there's a working model of this proposed approach somewhere, I'd like to give it a try if possible? We are currently running into this issue due to the use of `unicorn_worker_killer` and the number of files that end up getting created whenever a new worker process gets spun up. Over time the `/metrics` endpoint becomes very slow and memory usage increases unboundedly.

dmagliola:
Hi @saviogl, there isn't sadly one.
This is next on my "things to fix" list, and i'm hoping to get to it "in the next couple of weeks", but I don't want to under-deliver on that promise.

The way to try to solve this is to declare your own class that exposes the [FileStore interface](https://github.com/prometheus/client_ruby/tree/master/lib/prometheus/client/data_stores), and you'd point to that when declaring metrics.

The key thing to keep in mind here is that, while you'll have a lot of MetricStores, they all need to point to the same, shared instance of `FileMappedDict`, which you'd define at the FileStore level, not in the MetricStores as we do in `DirectFileStore`.

It's a non-trivial change, but that's basically what needs to happen.

The challenge for me is I wanted to offer the option: you can have file per process, or file per process-metric.
Unfortunately, you can't do that with a flag as this PR proposes (or not without making the code *a mess*), so we'll probably have to either choose one or the other to offer, or offer two separate DataStores to pick from, which have *a lot* of code duplication.

I'm sure there's a better way i'm missing, but anyway, that's my thinking on this.

saviogl:
@dmagliola Just trying to understand better whether or not this will actually solve our issue. While the amount of files is a problem to be had (maybe soon), the biggest issue right now is around memory consumption due to the fact we are "constantly" reaping unicorn worker processes for graceful memory management and therefore creating new "metrics" per processes. Moving to "One file per process instead of per metric/process for DirectFileStore" would tame the number of files created but the amount of data (metrics per process) would remain the same "correct"? And therefore we would still have an unbounded growth in regards to metrics -> memory for exporting those?

Is my assumption correct?

If I'm understanding the current model correctly, the solution for the memory unbounded growth would be to have processes to write the same file/metric with some process concurrent control system in place right?

Sinjo:
> While the amount of files is a problem to be had (maybe soon), the biggest issue right now is around memory consumption due to the fact we are "constantly" reaping unicorn worker processes for graceful memory management and therefore creating new "metrics" per processes.

If I've understood you right [this comment](https://github.com/prometheus/client_ruby/pull/214#issuecomment-803509081) in #214 is the situation running into, and you're generating a tonne of metric series by churning through unicorn workers. Whatever we come up with in there won't help you when it comes to scraping metrics becoming slow because of a large number of files, but might help you reduce the number of metrics (and ultimately timeseries in your Prometheus instance) produced.

wizardmatas:
FYI: on GitLab similar implementation is used: https://gitlab.com/gitlab-org/prometheus-client-mmap

PR #162, opened 2019-10-15
readme: remove warning of legacy pushgateway API

daveworth:
As of 0.9.0 (and in particular #102) Björn has made the warning of the legacy API no longer accurate.  Let's update the README too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,6 @@ The Ruby client can also be used to push its collected metrics to a
 where it's not possible or feasible to let a Prometheus server scrape a Ruby
 process. TLS and basic access authentication are supported.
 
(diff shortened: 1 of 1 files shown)


beorn7:
Please add the [DCO](https://github.com/prometheus/client_ruby/pull/162/checks?check_run_id=261292764).

I assume the Ruby client is actually not using the legacy API anymore? (If it did, it wouldn't work with recent versions of the PGW.)

beorn7:
Cool. Forgot about me own heroic deeds. :laughing: 

WRT CONTRIBUTING.md: Other Prometheus projects mention the DCO in their version of the file. I think it would make sense to pull it in here.

dmagliola:
Good catch! Thanks for this PR!

PR #165, opened 2019-10-28
Bump version to 0.11.0-alpha.1

dmagliola:
Bumping this to put it in production for a while. This is a very solid
candidate for a 1.0. If we don't find anything, we will release 1.0
next week

Signed-off-by: Daniel Magliola <danielmagliola@gocardless.com>

### DIFF
--- a/lib/prometheus/client/version.rb
+++ b/lib/prometheus/client/version.rb
@@ -2,6 +2,6 @@
 
 module Prometheus
   module Client
-    VERSION = '0.10.0'
+    VERSION = '0.11.0-alpha.1'
   end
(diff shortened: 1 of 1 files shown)


lib/prometheus/client/version.rb
```diff
@@ -2,6 +2,6 @@
 
 module Prometheus
   module Client
-    VERSION = '0.10.0'
+    VERSION = '0.11.0-aplha.1'
```
Sinjo:
```suggestion
    VERSION = '0.11.0-alpha.1'
```

PR #172, opened 2019-12-23
Add :most_recent aggregation to DirectFileStore

stefansundin:
Add `:most_recent` aggregation to DirectFileStore, which reports the value that was set by a process most recently.

I am using this for a small app called github-activity, and it tracks the remaining GitHub ratelimit using Prometheus (this is the only metric at the moment). The existing aggregations didn't really fit my use case, since for this use case, the last reported value from GitHub is the one that I care about. So I decided to try to improve the gem to support my use case, and I came up with this.

Please let me know about nitpicks and other things. I wasn't sure what the best name would be for this aggregation, and I was basically picking between `:latest` and `:most_recent`. And I am unsure if this should be using `CLOCK_MONOTONIC` or if `Time.now.to_f` is Ok.

Thanks!

Useful links:
- The GitHub repo for my app: https://github.com/stefansundin/github-activity
- The commit in question: https://github.com/stefansundin/github-activity/commit/936d4d6f1627549390fdf151ab475c448dde12cc
- The app is currently running at: https://gh-rss.herokuapp.com/
- The metrics endpoint is public here: https://gh-rss.herokuapp.com/metrics.

### DIFF
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -29,7 +29,7 @@ module Prometheus
 
       class DirectFileStore
         class InvalidStoreSettingsError < StandardError; end
-        AGGREGATION_MODES = [MAX = :max, MIN = :min, SUM = :sum, ALL = :all]
+        AGGREGATION_MODES = [MAX = :max, MIN = :min, SUM = :sum, ALL = :all, MOST_RECENT = :most_recent]
         DEFAULT_METRIC_SETTINGS = { aggregation: SUM }
         DEFAULT_GAUGE_SETTINGS = { aggregation: ALL }
 
@@ -121,7 +121,7 @@ module Prometheus
             stores_for_metric.each do |file_path|
               begin
                 store = FileMappedDict.new(file_path, true)
-                store.all_values.each do |(labelset_qs, v)|
+                store.all_values.each do |(labelset_qs, v, ts)|
                   # Labels come as a query string, and CGI::parse returns arrays for each key
                   # "foo=bar&x=y" => { "foo" => ["bar"], "x" => ["y"] }
                   # Turn the keys back into symbols, and remove the arrays
@@ -129,7 +129,7 @@ module Prometheus
                     [k.to_sym, vs.first]
                   end.to_h
 
-                  stores_data[label_set] << v
+                  stores_data[label_set] << [v, ts]
                 end
               ensure
                 store.close if store
@@ -181,30 +181,41 @@ module Prometheus
           end
 
           def aggregate_values(values)
-            if @values_aggregation_mode == SUM
-              values.inject { |sum, element| sum + element }
-            elsif @values_aggregation_mode == MAX
-              values.max
-            elsif @values_aggregation_mode == MIN
-              values.min
-            elsif @values_aggregation_mode == ALL
-              values.first
+            # Each entry in the `values` array is a tuple of `value` and `timestamp`,
+            # so for all aggregations except `MOST_RECENT`, we need to only take the
+            # first value in each entry and ignore the second.
+            if @values_aggregation_mode == MOST_RECENT
+              latest_tuple = values.max { |a,b| a[1] <=> b[1] }
+              latest_tuple.first # return the value without the timestamp
             else
-              raise InvalidStoreSettingsError,
-                    "Invalid Aggregation Mode: #{ @values_aggregation_mode }"
+              values = values.map(&:first) # Discard timestamps
+
+              if @values_aggregation_mode == SUM
+                values.inject { |sum, element| sum + element }
+              elsif @values_aggregation_mode == MAX
+                values.max
+              elsif @values_aggregation_mode == MIN
+                values.min
+              elsif @values_aggregation_mode == ALL
+                values.first
+              else
+                raise InvalidStoreSettingsError,
+                      "Invalid Aggregation Mode: #{ @values_aggregation_mode }"
+              end
             end
           end
         end
 
         private_constant :MetricStore
 
-        # A dict of doubles, backed by an file we access directly a a byte array.
+        # A dict of doubles, backed by an file we access directly as a byte array.
         #
         # The file starts with a 4 byte int, indicating how much of it is used.
         # Then 4 bytes of padding.
         # There's then a number of entries, consisting of a 4 byte int which is the
         # size of the next field, a utf-8 encoded string key, padding to an 8 byte
-        # alignment, and then a 8 byte float which is the value.
+        # alignment, and then a 8 byte float which is the value, and then a 8 byte
+        # float which is the unix timestamp when the value was set.
         class FileMappedDict
           INITIAL_FILE_SIZE = 1024*1024
 
@@ -236,7 +247,8 @@ module Prometheus
               @positions.map do |key, pos|
                 @f.seek(pos)
                 value = @f.read(8).unpack('d')[0]
-                [key, value]
+                timestamp = @f.read(8).unpack('d')[0]
+                [key, value, timestamp]
               end
             end
           end
@@ -258,7 +270,7 @@ module Prometheus
 
             pos = @positions[key]
             @f.seek(pos)
-            @f.write([value].pack('d'))
+            @f.write([value, Time.now.to_f].pack('dd'))
             @f.flush
           end
 
@@ -299,7 +311,7 @@ module Prometheus
           def init_value(key)
             # Pad to be 8-byte aligned.
             padded = key + (' ' * (8 - (key.length + 4) % 8))
-            value = [padded.length, padded, 0.0].pack("lA#{padded.length}d")
+            value = [padded.length, padded, 0.0, 0.0].pack("lA#{padded.length}dd")
             while @used + value.length > @capacity
               @capacity *= 2
               resize_file(@capacity)
@@ -310,7 +322,7 @@ module Prometheus
             @f.seek(0)
             @f.write([@used].pack('l'))
             @f.flush
-            @positions[key] = @used - 8
+            @positions[key] = @used - 16
           end
 
           # Read position of all keys. No locking is performed.
@@ -320,7 +332,7 @@ module Prometheus
               padded_len = @f.read(4).unpack('l')[0]
               key = @f.read(padded_len).unpack("A#{padded_len}")[0].strip
               @positions[key] = @f.pos
-              @f.seek(8, :CUR)
+              @f.seek(16, :CUR)
             end
           end
         end
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -45,7 +45,7 @@ module Prometheus
           end
 
           settings = default_settings.merge(metric_settings)
-          validate_metric_settings(settings)
+          validate_metric_settings(metric_type, settings)
 
           MetricStore.new(metric_name: metric_name,
                           store_settings: @store_settings,
@@ -54,7 +54,7 @@ module Prometheus
 
         private
 
-        def validate_metric_settings(metric_settings)
+        def validate_metric_settings(metric_type, metric_settings)
           unless metric_settings.has_key?(:aggregation) &&
             AGGREGATION_MODES.include?(metric_settings[:aggregation])
             raise InvalidStoreSettingsError,
@@ -64,6 +64,11 @@ module Prometheus
           unless (metric_settings.keys - [:aggregation]).empty?
             raise InvalidStoreSettingsError,
                   "Only :aggregation setting can be specified"
+          end
+
+          if metric_settings[:aggregation] == MOST_RECENT && metric_type != :gauge
+            raise InvalidStoreSettingsError,
+                  "Only :gauge metrics support :most_recent aggregation"
           end
         end
 
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -105,6 +105,12 @@ module Prometheus
           end
 
           def increment(labels:, by: 1)
+            if @values_aggregation_mode == DirectFileStore::MOST_RECENT
+              raise InvalidStoreSettingsError,
+                    "The :most_recent aggregation does not support the use of increment"\
+                      "/decrement"
+            end
+
             key = store_key(labels)
             in_process_sync do
               value = internal_store.read_value(key)
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -279,9 +279,10 @@ module Prometheus
               init_value(key)
             end
 
+            now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
             pos = @positions[key]
             @f.seek(pos)
-            @f.write([value, Time.now.to_f].pack('dd'))
+            @f.write([value, now].pack('dd'))
             @f.flush
           end
 
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -257,8 +257,7 @@ module Prometheus
             with_file_lock do
               @positions.map do |key, pos|
                 @f.seek(pos)
-                value = @f.read(8).unpack('d')[0]
-                timestamp = @f.read(8).unpack('d')[0]
+                value, timestamp = @f.read(16).unpack('dd')
                 [key, value, timestamp]
               end
             end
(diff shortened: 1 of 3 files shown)


stefansundin:
Hi @SuperQ @Sinjo @dmagliola. Feel free to let me know what you think. Thanks.

SuperQ:
Thinking about the comparison of gauge values are handled in other languages, and when in single-threaded mode, "most recent" is the standard way to do things. I wonder if this should be the default way multi-process gauges work.

brian-brazil:
> I wonder if this should be the default way multi-process gauges work.

It requires cross-process locking, and we're trying to avoid that on performance grounds.

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -68,6 +68,9 @@ def validate_metric_settings(metric_settings)
         end
 
         class MetricStore
+          # Update the filename prefix when the file format changes
+          FILENAME_PREFIX = "metricv2"
```
dmagliola: Thought: should the filename_prefix be the `Prometheus::Client::VERSION`? That way, we get incrementing of this value for free, without having to remember to update it.

Another thought: we're already working on the assumption that you start with a clear `/tmp`, so this feels a little out of place by itself, given that a server restart where you keep your old files is already a bad thing and we're not really helping with that by doing this.

I think i'd prefer to not have this second change

dmagliola:
First of all, thank you for this PR! This is a great contribution :-D
So, I have a number of thoughts on this. Not necessarily in order... 

- I like the idea, since it's something we were thinking of doing for our own purposes. @lawrencejones is this more or less what you had in mind? Are you happy with this approach?

- Also @lawrencejones  or @Sinjo do you have an opinion on  `CLOCK_MONOTONIC` vs `Time.now.to_f`? You both have a clearer idea of this than I do.

- I was a little concerned about the performance of this, but it seems to be roughly the same as what we currently have (2% to 3% slower, maybe), so 👍 on that. Files are probably going to be a bit bigger, but it doesn't seem to be a huge deal.

- @SuperQ I kinda like the idea of making this the default, because our current default (`all`) creates a lot of time series, potentially unnecessarily. My main worry is that that's a backwards-incompatible change, so i'm not sure we can introduce that. Maybe we should start a list of breaking changes that we'd like to introduce next time we release a major version, and start with that one?

- @brian-brazil Unless i'm misunderstanding your comment, this shouldn't be a problem the way this is implemented. This is not locking across processes. Instead, each value is paired with the timestamp of when it was set, and which one was latest gets resolved at scrape time, which I thought was quite an elegant approach. Does that alleviate that concern, or did I misunderstand?

- @stefansundin The thing that gives me pause is that, in the cases where we wanted this feature, we got around it by either doing a `min` aggregation on the data store, or on the Prometheus Query. In your case, it feels like this would do the trick too, right? The "latest" and the "minimum" should be the same given that it's a rate limit that presumably decreases over time?  In any case, you'd need the latest value *across* all your servers, which this wouldn't be able to give you, so you'd still need the aggregation at the query level. With that in mind, do we need this feature?

Finally, Merry Xmas to all of you :-)

lib/prometheus/client/data_stores/direct_file_store.rb
stefansundin: I had both thoughts too. However, my conclusion was that since the gem isn't cleaning up these files on its own, the burden is currently on the app to do this. I would bet that someone will run into this problem.

See this part of my app: https://github.com/stefansundin/github-activity/commit/936d4d6f1627549390fdf151ab475c448dde12cc#diff-d21913acbe4d1528a62c64cdec6d43d7R11-R14

After this PR, I was going to submit another PR that adds a new flag to DirectFileStore, called `:clean_dir` or something similar, which if true would clean all the directory for you.

A little bit related, that PR would also contain a new method that you can call with a pid, and it would clean up any metrics that used to belong to that process. I would replace this code with a call to that method: https://github.com/stefansundin/github-activity/commit/936d4d6f1627549390fdf151ab475c448dde12cc#diff-3f0c524806c4374c99a509efaffd3e8aR29-R32

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -182,13 +185,15 @@ def process_id
 
           def aggregate_values(values)
             if @values_aggregation_mode == SUM
-              values.inject { |sum, element| sum + element }
+              values.map { |element| element[0] }.inject { |sum, value| sum + value }
```
dmagliola: For all 3 of these, could we `map(&:first)` and then do the same thing we were doing before?
That feels a lot more readable to me.

(so: `values.map(&:first).inject { |sum, element| sum + element }`, `values.map(&:first).max`, etc)
I'm half wondering why we did this `inject` before, instead of just calling `.sum`... Is there a large performance penalty to `map`? Could we measure that?
The reason I mention that is that we already have some users with performance problems on scrape, and i'd thus like to make this as fast as possible.

If `map` is much slower, we should probably change this first one to the same inject as before, but the body of the block to `sum + element[0]`...
And i'm not sure about the ones below, whether that's the fastest way?

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -182,13 +185,15 @@ def process_id
 
           def aggregate_values(values)
```
dmagliola: Could we also add a comment at the top of the method explaining that each entry in the `values` array is a tuple of "value, timestamp"? That makes the rest of the code a lot easier to follow

stefansundin:
> First of all, thank you for this PR! This is a great contribution :-D

That's a great thing to hear.

> * @stefansundin The thing that gives me pause is that, in the cases where we wanted this feature, we got around it by either doing a `min` aggregation on the data store, or on the Prometheus Query. In your case, it feels like this would do the trick too, right? The "latest" and the "minimum" should be the same given that it's a rate limit that presumably decreases over time?  In any case, you'd need the latest value _across_ all your servers, which this wouldn't be able to give you, so you'd still need the aggregation at the query level. With that in mind, do we need this feature?

So my ratelimit starts at 5000 requests/hour, and decreases, that is correct. But the ratelimit is reset every hour, so if I used `min` then I'd always see the lowest value even after the reset. If I have multiple servers, they would probably have their own request IP from GitHub's point of view, and in this case the ratelimit would be independent, so I would need to track all of them individually. This change lets me omit metric series that I already know are "wrong" (in the sense that I already know, at scrape-time, that the information is out of date).

> Finally, Merry Xmas to all of you :-)

Merry Christmas to all of you too! :)

lib/prometheus/client/data_stores/direct_file_store.rb
stefansundin: I am very open to changing this code. I mostly just wanted to get it working without considering performance that much.

brian-brazil:
> This is not locking across processes. Instead, each value is paired with the timestamp of when it was set, and which one was latest gets resolved at scrape time, which I thought was quite an elegant approach.

I hadn't looked at the code, I was responding to @SuperQ's comment. The issue with this approach is that it only works for set, not inc/dec. Thus it cannot be the default.

lib/prometheus/client/data_stores/direct_file_store.rb
stefansundin: I opened a second PR, so I can drop this commit from this PR. https://github.com/prometheus/client_ruby/pull/173

lawrencejones:
I think banning inc/dec on this aggregation is probably the way to go. Hell, we could raise for the moment on inc/dec while allowing set, then add the inter-process locking for inc/dec later if people really want it.

brian-brazil:
If you want inc/dec then there's no need for inter-process coordination, each process can keep it locally and we sum on exposition.

dmagliola:
That's a really good point on `inc/dec`, Brian, thanks for raising that.

@lawrencejones I like your solution on first sight, but i'm a bit worried of two things:
- the store behaves somewhat differently when *observing* based on the aggregation method you choose, which feels a bit weird to me.
- more importantly, i'm a bit hesitant about raising an exception, since it's highly possible you won't find out about it until it's running in production. I can very easily see how you'd use the in-memory stores for dev/test, and only use the file store in prod. (And the reason this distinction is important is that the only store that supports aggregation is the FileStore, the others wouldn't know they should raise)

Side thought... Would / could this be solved by "Custom Collectors", instead?
I'm thinking of the value to be exposed getting calculated when the scrape happens, instead of storing separate values for each process, and then having to resolve which one is the latest...

Sinjo:
> do you have an opinion on `CLOCK_MONOTONIC` vs `Time.now.to_f`? You both have a clearer idea of this than I do.

I think either is fine for this use-case. Each has their caveats, but I think `CLOCK_MONOTONIC` is ever so slightly better for what we're doing and I've included my reasoning below.

The caveat I had to read up on with `CLOCK_MONOTONIC` was whether it was consistent between different processes. It turns out that as long as you're comparing values from the same boot of the same computer, you're fine. If you compare values from different computers or from different boots of the same computer, they will be meaningless.

`Time.now.to_f` will be susceptible to clock adjustments by ntp, but not to DST changes. Just tested with two `irb` sessions, leaving one in London time and setting `TZ=Europe/Paris` in the env of the other. There's no 1 hour offset between them (unlike `Time.now` by itself).

dmagliola:
Hello @stefansundin,
Sorry about the delay in responding.

So, I think we should merge this feature, but i'd like to ask for a few changes, because as it stands, it's quite possible to use this in a way that will do something the user definitely didn't mean.

Could we have:
- a validation in `validate_metric_settings` that makes sure `most_recent` can only be used for Gauges? Using this aggregation with any other metric type is almost certainly not what the user intended, and it'll result in counters that go up and down, and completely inconsistent histograms.
- a validation on `increment` that throws an exception if the aggregation type is most_recent. This will also result in not what the user wants.
- documentation in the `README` of this aggregation mode, when it's useful, and a big note that it's only useful for Gauges, and that it doesn't allow using `increment/decrement`
- i'd get rid of the filename prefix / renaming for this PR. The code so far assumes you're starting from an empty directory, and having that in this PR mixes the concerns we're trying to solve.
 - see my comment on adding a comment above `aggregate_values` documenting what it returns, which should make it much easier to follow the code. Also the code about mapping / how to aggregate the values.
- let's switch to `CLOCK_MONOTONIC` as per @Sinjo's comment above.

Notes on making this the default:
- SInce this has unobvious caveats, like the fact that you can't use inc/dec on gauges that have this aggregation, and that you might only find about that in production if your test environment doesn't use the FileStore, I would like to make it a manual option that someone has to find out about by reading the docs, at which point we warn them about this. If this is the default, we really need to make inc/dec work, which is quite a lot of work if we don't want to be locking between processes.

Thank you!

lib/prometheus/client/data_stores/direct_file_store.rb
stefansundin: This commit has been dropped.

stefansundin:
@dmagliola I dropped the commit with the filename prefix change. I'd highly recommend testing the behavior for loading files with the old format. If there isn't any work to prevent it from happening, then I will bet that you will get reports of crashes (or whatever the behavior will be) from confused users. Hopefully only from dev machines and not from production.

As for the rest, unfortunately I am short on time to work on this at the moment. Can you work on addressing the rest of the changes? This branch is allowing edits from maintainers, so if you want you can push new commits to it and then merge when desired. Thanks.

dmagliola:
Alright, i've made all those changes mentioned above. It's now pending review and we'll merge this and cut a new minor version

lib/prometheus/client/data_stores/direct_file_store.rb
```diff
@@ -181,30 +192,41 @@ def process_id
...
+            # Each entry in the `values` array is a tuple of `value` and `timestamp`,
+            # so for all aggregations except `MOST_RECENT`, we need to only take the
+            # first value in each entry and ignore the second.
+            if @values_aggregation_mode == MOST_RECENT
+              latest_tuple = values.max { |a,b| a[1] <=> b[1] }
+              latest_tuple.first # return the value without the timestamp
             else
-              raise InvalidStoreSettingsError,
-                    "Invalid Aggregation Mode: #{ @values_aggregation_mode }"
+              values = values.map(&:first) # Discard timestamps
```
Sinjo, line 202: Not sure whether we should block the PR on this, but might we be slowing down the metric read path here with extra copying?

spec/prometheus/client/data_stores/direct_file_store_spec.rb
```diff
@@ -32,6 +32,31 @@
                          metric_type: :counter,
                          metric_settings: { some_setting: true })
     end.to raise_error(Prometheus::Client::DataStores::DirectFileStore::InvalidStoreSettingsError)
+
+    # :most_recent aggregation can only be used for gauges
```
Sinjo, line 36: I'd prefer this to be done with RSpec methods describing the tests rather than a single comment at the top of a bunch of `expect`s.

spec/prometheus/client/data_stores/direct_file_store_spec.rb
```diff
@@ -267,6 +292,57 @@
...
+      metric_store1.set(labels: { foo: "bar" }, val: 1)
+      allow(Process).to receive(:pid).and_return(23456)
+      metric_store2.set(labels: { foo: "bar" }, val: 3)
+      allow(Process).to receive(:pid).and_return(12345)
+      metric_store1.set(labels: { foo: "baz" }, val: 7)
+      allow(Process).to receive(:pid).and_return(23456)
+      metric_store2.set(labels: { foo: "baz" }, val: 2)
+      allow(Process).to receive(:pid).and_return(12345)
+      metric_store1.set(labels: { foo: "baz" }, val: 4)
+      allow(Process).to receive(:pid).and_return(23456)
```
Sinjo, line 318: Not sure what these last two examples are here to test. I think all the behaviour is covered by the ones above?

spec/prometheus/client/data_stores/direct_file_store_spec.rb
```diff
@@ -267,6 +292,57 @@
...
+        metric_type: :gauge,
+        metric_settings: { aggregation: :most_recent }
+      )
+      metric_store2 = subject.for_metric(
+        :metric_name,
+        metric_type: :gauge,
+        metric_settings: { aggregation: :most_recent }
+      )
+
+      allow(Process).to receive(:pid).and_return(12345)
```
Sinjo, line 308: I find this block kinda dense to read. Wondering if either [RSpec multiple return](https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses/returning-a-value#specify-different-return-values-for-multiple-calls) syntax or just breaking it up with newlines would help.

I think the fact that the lines are all identical length somehow makes it harder to read.

lib/prometheus/client/data_stores/direct_file_store.rb
dmagliola: Possibly. Since we've done practically no optimization whatsoever on the read path, however, and we don't have any real benchmarks, i'd say let's focus on that when we work on that performance, rather than here.

I totally agree with "at leats not making it worse", but in this case, I can't think of a clear, easy alternative that still works with calling `max`, `min`, etc

spec/prometheus/client/data_stores/direct_file_store_spec.rb
dmagliola: i'm not really sure what shape the code you're thinking of has, sorry.
to me this is doing that, if we remove the comment, this is pretty much how the rest of the test case is written.

Can you mock a quick example of what you mean by "done with RSpec methods"?

spec/prometheus/client/data_stores/direct_file_store_spec.rb
dmagliola: maybe i should leave only one of those two. What this is testing is a labelset that only exists in one of the stores, all the others are in both.

spec/prometheus/client/data_stores/direct_file_store_spec.rb
dmagliola: So this is one of the challenges of coming back to code 4 months later... I don't remember why this is written this way...

I'm not really sure why we have lines 312 and 313, for example...
I'm also not entirely sure why we need to be going back and forth with the process IDs. i think it's so the "pid change detector" doesn't change the file you're reading... let me take a stab at rewriting this...

dmagliola: @Sinjo I've just reordered and separated the lines, hopefully it'll make a bit more sense (check the last commit, the fixup one)

lib/prometheus/client/data_stores/direct_file_store.rb
Sinjo: Let's focus on this when we come to optimising the read path then.

spec/prometheus/client/data_stores/direct_file_store_spec.rb
Sinjo: LGTM

PR #175, opened 2019-12-29
Change prometheus command to use double-dash for flags

chrohrer:
This PR replaces the pre-2.0 style flag prefix (single dash) with the current double-dash flag prefix for the prometheus command in examples/rack/README.md

@dmagliola


dmagliola:
Good catch, thank you!

PR #176, opened 2020-01-08
observed value should be stored the bucket

ganmacs:
whose value is the same as the observed value

Seems to be working  in v0.9.0 
https://github.com/prometheus/client_ruby/blob/9bae72520e7fdadab983ddd2cd40832e37489793/lib/prometheus/client/histogram.rb#L29

which behavior is correct?

### DIFF
--- a/lib/prometheus/client/histogram.rb
+++ b/lib/prometheus/client/histogram.rb
@@ -47,7 +47,7 @@ module Prometheus
       end
 
       def observe(value, labels: {})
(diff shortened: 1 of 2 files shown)


dmagliola:
Hello @ganmacs. Sorry for the very late reply!
Thank you for sending this fix.
While this is a fix, this is a breaking change, so we'll be cutting a new version of the API soon with this included.
Sorry again it took so long to get back

PR #177, opened 2020-01-14
Histogram bucket helpers

lawrencejones:
Most client libraries provide helpers for generating linear/exponential
histogram buckets. This provides the same interface as the golang
client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,11 @@ histogram.get(labels: { service: 'users' })
 # => { 0.005 => 3, 0.01 => 15, 0.025 => 18, ..., 2.5 => 42, 5 => 42, 10 = >42 }
 ```
 
+Histograms provide default buckets of `[0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10]`
+
+You can specify your own buckets, either explicitly, or using the `Histogram.linear_buckets`
+or `Histogram.exponential_buckets` methods to define regularly spaced buckets.
+
 ### Summary
 
 Summary, similar to histograms, is an accumulator for samples. It captures
--- a/lib/prometheus/client/histogram.rb
+++ b/lib/prometheus/client/histogram.rb
@@ -33,6 +33,14 @@ module Prometheus
               store_settings: store_settings)
       end
 
+      def self.linear_buckets(start:, width:, count:)
+        count.times.map { |idx| start.to_f + idx * width }
+      end
+
+      def self.exponential_buckets(start:, factor: 2, count:)
+        count.times.map { |idx| start.to_f * factor ** idx }
+      end
+
       def with_labels(labels)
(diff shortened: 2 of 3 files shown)


lawrencejones:
FYI @dmagliola

lawrencejones:
Any love @Sinjo?

lib/prometheus/client.rb
```diff
@@ -14,5 +14,13 @@ def self.registry
...
       @config ||= Config.new
     end
+
+    def self.linear_buckets(start:, width:, count:)
+      count.times.map { |idx| start.to_f + idx * width }
+    end
+
+    def self.exponential_buckets(start:, factor: 2, count:)
+      count.times.map { |idx| start.to_f * factor ** idx }
+    end
```
dmagliola: I love this feature, it was actually on our TO-DO list, so glad you tackled it.
I'm not sure about the location of these two methods, though...

Could you move these to be static methods in the `Histogram` class?

Also, should we validate the parameters, like the Go client does?

lawrencejones: Perhaps validating is an idea? Will have a look at that.

On the class method part, is it worth moving them there? Bearing in mind that would be different from the other clients.

dmagliola: I think it's worth moving.
That's where I would look for them as a user of this library. 
And unless I misread the Go code, that's where they are in the Go client too.

Sinjo:
Guess who forgot to specifically set up notifications for all activity on this project 🤦🏻‍♂️

Didn't realise when we were working super actively on 1.0 and only thought to go looking after you tagged me! Gonna look over this now.

lib/prometheus/client.rb
dmagliola: Ah, I did misread the Go code. I saw it in Histogram.go and assumed it'd be there.
I think the conventions in these two languages are different.
Whereas you do `NewHistogram` in Go, you go `Histogram.new` in Ruby. Everything Histogram related is in the Histogram class itself.
The Client class is only used once for setup on startup, or not at all. It doesn't look like a place where someone would look when creating a histogram.

On that note, though... Could we also add a little section in the `README` documenting these two methods? :D

Sinjo: Having them on `Histogram` would definitely match my expectations when writing Ruby. I think going with typical idiom within the language is a very fair shout. In `client_java` they're [chainable methods](https://github.com/prometheus/client_java/blob/ef16c43f63128c14627eeff301a32c12e7751a4e/simpleclient/src/main/java/io/prometheus/client/Histogram.java#L59) on the Histogram builder.

dmagliola:
Ok, i've made the changes I mentioned above (moving the helper methods to the `Histogram` class, and added a mention of the feature in the README.
@Sinjo can you take a look?

OsvaldoRino (approved):

PR #180, opened 2020-02-06
Add :labels for Prometheus::Middleware::Collector

benyitzhaki:
This pull request makes it possible to pass a labels argument that will add default labels to each metric exposed using the collector middleware.
Usage example:
` 
config.middleware.use Prometheus::Middleware::Collector, { labels: { env: Rails.env } }
`

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -1,8 +1,6 @@
 # encoding: UTF-8
-
 require 'benchmark'
 require 'prometheus/client'
-
 module Prometheus
   module Middleware
     # Collector is a Rack middleware that provides a sample implementation of a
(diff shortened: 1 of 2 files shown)


brian-brazil:
That sounds like a target label, which you're best off adding on the
Prometheus end.

Sinjo:
Hi @benyitzhaki,

First off, thanks for taking time to send us a PR.

I agree with @brian-brazil on this one. What's worth mentioning is that we did have a [more flexible version of this that we removed pre-1.0](https://github.com/prometheus/client_ruby/pull/121). The reason we removed it wasn't that we disliked the feature, but that we felt the API was confusing. We would potentially be open to bringing that version back if the concerns in #111 were addressed.

I'm going to close this PR, but we'd be happy to hear suggestions on a more intuitive API for the version we removed in #121.

PR #199, opened 2020-07-30
Add port option to exporter middleware

tragiclifestories:
If the port option is set, all requests for /metrics on other ports will be forwarded to the app. If it is unset or nil, or the ports match, export things as usual.

Allows separate mounting of metrics and main app to enforce different security setups etc.

### DIFF
--- a/lib/prometheus/middleware/exporter.rb
+++ b/lib/prometheus/middleware/exporter.rb
@@ -21,11 +21,12 @@ module Prometheus
         @app = app
         @registry = options[:registry] || Client.registry
         @path = options[:path] || '/metrics'
+        @port = options[:port]
         @acceptable = build_dictionary(FORMATS, FALLBACK)
       end
 
       def call(env)
-        if env['PATH_INFO'] == @path
+        if metrics_port?(env['SERVER_PORT']) && env['PATH_INFO'] == @path
           format = negotiate(env, @acceptable)
           format ? respond_with(format) : not_acceptable(FORMATS)
         else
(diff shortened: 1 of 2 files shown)


dmagliola:
I've been thinking about this again, and while my posture has always been that we shouldn't permit much configurability of these middlewares, encouraging implementers to use these as a template and make their own version instead...

I have realized that in that "policy" I've been conflating both middlewares unfairly.
I still think that we shouldn't allow much configurability of the  the **Collector** middleware, as it gets messy quite quickly, and the changes *will* be quite specific to each user of the client.

However, I no longer think this should apply to the **Exporter**. Or at least it doesn't need to as much.
This change is a great example of adding quite useful functionality, that probably many users will need, without really adding much complexity. I think we should embrace these.

I'm going to go through other closed PRs to see if this new thinking applied to any of them, and I closed them when I maybe shouldn't have.

I'm sorry @tragiclifestories that we took forever to respond to this. This sort of landed in the middle of a wider conversation we've been having for a while, and it got bundled with a whole category of changes that I now realize it doesn't really belong with. My bad!

If you rebase this to add the "signoff", I'd be happy to merge this.
Not sure if @Sinjo has a different opinion.

dmagliola (approved):

tragiclifestories:
Fuck, I'd completely forgotten about this. Will do in the new year. Merry Xmas!

tragiclifestories:
Had 5 spare minutes and a hangover, so done this now.

Sinjo (approved):
+1 on this. Let's get it in and add it to the docs in a subsequent PR.

PR #201, opened 2020-08-25
Make all registry methods thread safe

matthieuprat:
The registry is backed by a Hash, which is not guaranteed to be thread
safe on all interpreters. For peace of mind, this change synchronizes
all accesses to the metrics hash.

Another option would have been to use a [thread-safe Hash][1] instead of
a Hash but this would have meant adding Ruby Concurrent as a
dependency, which I'm assuming we don't want.

Ref: https://github.com/prometheus/client_ruby/pull/184#discussion_r406038191

[1]: https://github.com/ruby-concurrency/concurrent-ruby/blob/v1.1.7/lib/concurrent-ruby/concurrent/hash.rb

### DIFF
--- a/lib/prometheus/client/registry.rb
+++ b/lib/prometheus/client/registry.rb
@@ -22,7 +22,7 @@ module Prometheus
         name = metric.name
 
         @mutex.synchronize do
-          if exist?(name.to_sym)
+          if @metrics.key?(name.to_sym)
             raise AlreadyRegisteredError, "#{name} has already been registered"
           end
           @metrics[name.to_sym] = metric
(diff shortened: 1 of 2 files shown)


lib/prometheus/client/registry.rb
```diff
@@ -22,7 +22,7 @@ def register(metric)
         name = metric.name
 
         @mutex.synchronize do
-          if exist?(name.to_sym)
+          if @metrics.key?(name.to_sym)
```
matthieuprat, line 25: I made this change because `Mutex` is not re-entrant. I guess we could use a `Monitor` instead but inlining the `exist?` method actually makes the code slightly more readable in my opinion.

dmagliola (approved):

dmagliola:
Sorry I somehow missed this for a month and a half!

PR #204, opened 2020-10-02
Takes into account the SCRIPT_NAME in path

woto:
Hi! Not sure that this is correct, but: 
1. We have lost the path using mountable engines (twirp, grape) 
2. It looks like the RFC says that this is correct: [SCRIPT_NAME](https://tools.ietf.org/html/rfc3875#section-4.1.13), [PATH_NFO](https://tools.ietf.org/html/rfc3875#section-4.1.5)
3. https://github.com/rack/rack/blob/master/lib/rack/handler/webrick.rb#L92

Sorry without tests for now. Want to hear your opinion.

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -67,15 +67,16 @@ module Prometheus
       end
 
       def record(env, code, duration)
+        path = [env["SCRIPT_NAME"], env['PATH_INFO']].join
         counter_labels = {
           code:   code,
           method: env['REQUEST_METHOD'].downcase,
-          path:   strip_ids_from_path(env['PATH_INFO']),
+          path:   strip_ids_from_path(path),
         }
 
         duration_labels = {
           method: env['REQUEST_METHOD'].downcase,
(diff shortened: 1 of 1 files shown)


dmagliola:
Hi @woto,

Thank you for your PR.
This is the same issue from PR #197 

Our concern with this is that it feels like it's a backwards incompatible change, and our recommendation at the time is to use this built-in collector as an "example" that you are encouraged to use as a base to make your own if it isn't fit for purpose.

I'd like to close the PR and move the conversation to #197 if that's ok, so we can have it in one place.
Thanks!

woto:
Ok, fine thank you

dmagliola:
Ok, we've finally had time to look into this enough to convince ourselves this is definitely the right thing to do.
I wanted to add tests to this before merging, but I wasn't able to in this same branch, so I made a new PR #206 
I've added you as a co-author on the commit, to give credit for the code and idea, hope that's ok!

I'm going to merge #206 son, and cut a new major version soon that fixes this.

Thank you for bringing this to our attention and for the PR!

PR #206, opened 2020-10-14
Include SCRIPT_NAME when determining path in Collector

dmagliola:
When determining the path for a request, `Rack::Request` prefixes the
`SCRIPT_NAME`, [as seen here][1].

This is a problem with our current code when using mountable engines,
where the engine part of the path gets lost.

This patch fixes that to include `SCRIPT_NAME` as part of the path.

NOTE: This is not backwards compatible. Labels will change in existing
metrics. We will cut a new major version once we ship this.

[1]: https://github.com/rack/rack/blob/294fd239a71aab805877790f0a92ee3c72e67d79/lib/rack/request.rb#L512

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -67,15 +67,17 @@ module Prometheus
       end
 
       def record(env, code, duration)
+        path = [env["SCRIPT_NAME"], env['PATH_INFO']].join
+
         counter_labels = {
           code:   code,
           method: env['REQUEST_METHOD'].downcase,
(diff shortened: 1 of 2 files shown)


dmagliola:
@ianks @woto Thank you for your PRs! I couldn't push tests into your branches, so made this new PR and tagged you as co-authors in the commit. I hope that's ok with you.
Hoping to merge this soon

Sinjo (approved):

ianks:
Awesome stuff. Excited for the new major!

Issue #208, opened 2020-12-02
Client should automatically initialize label-less metrics to 0


dmagliola:
We provide the `init_label_set` method to deal with the "Missing Metrics" problem on metrics that have labels.
However, we should automatically initialize the empty labelset for metrics that don't take any labels, so a 0 is exported.

A caller _could_ call `init_label_set({})`, but it's unintuitive, and the [Prometheus documentation itself](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics) sets the expectation that this will be done for you.

**Q: Is this a patch, minor or major?**
On the one hand, this is kind of a bug fix. We should've done this when we introduced `init_label_set`
On the other hand, this is definitely going to affect the output of a client, which may surprise some folk, so it may count as a breaking change. But... It's making that output "more correct". I'm not entirely sure here.

**Q: We should 100% do this for counters. Should we also do it for Gauges, etc?**
I'd take a cue from the Python/Go clients, do the same they do.
NB: If i'm reading the code correctly, the Python library does this, by calling `_metric_init` for `is_observable` metrics, which includes label-less ones, and those with pre-set label values.

Sinjo:
Fixed by #209

PR #209, opened 2020-12-24
Automatically initialize time series without labels

dmagliola:
Issue #208 

According to [Prometheus Best Practices](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics), client libraries are expected to automatically export a 0 value when declaring a metric that has no labels.

We missed this so far in the Ruby client, this PR rectifies that.

NOTE: This can be considered a breaking change. On the one hand, it's a bug fix, but on the other, it will make many time series materialize when scraping apps that use the client, that previously wouldn't have.

Depending on how many label-less metrics the app is declaring, this may have a significant impact, so we should probably cut a new major version and warn about this in the Release Notes.

### DIFF
--- a/lib/prometheus/client/metric.rb
+++ b/lib/prometheus/client/metric.rb
@@ -40,6 +40,8 @@ module Prometheus
           metric_type: type,
           metric_settings: store_settings
         )
+
+        init_label_set({}) if labels.empty?
       end
 
       # Returns the value for the given label set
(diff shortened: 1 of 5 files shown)


Sinjo (approved):
I'm 👍🏻 on this, and I think we should go with a major bump as a courtesy. Hitting capacity on your Prometheus database is not a fun experience (often friends have reported just dropping all their metric history when this happens, which is kinda sad). If we can avoid ruining people's day like that for some thing as cheap as bumping the major version number, we should.

Sinjo:
I've added this to a v3.0.0 milestone. If we have any other changes (either outstanding PRs or issues that won't take us much work - so probably not OpenMetrics) we think need the major version bump we should bundle them in there.

PR #210, opened 2020-12-25
Support latest Ruby versions

dmagliola:
Ruby 3 was released today.
And the good news is: we are already perfectly compatible with it.
`rspec` runs with 0 warnings, which was a surprise.


SuperQ (approved):
Nice

PR #211, opened 2021-01-01
let signed_id of ActiveStorge::Blob be striped from as an id

rzjfr:
Strip signed_id of [Rails ActiveStorge::Blob](https://api.rubyonrails.org/v6.1.0/classes/ActiveStorage/Blob.html) by default like normal IDs and UUIDs

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -91,6 +91,7 @@ module Prometheus
         path
           .gsub(%r{/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(/|$)}, '/:uuid\\1')
           .gsub(%r{/\d+(/|$)}, '/:id\\1')
+          .gsub(%r{/[0-9a-zA-Z=]+--[0-9a-f]+(/|$)}, '/:signed_id\\1')
       end
     end
   end
--- a/spec/prometheus/middleware/collector_spec.rb
+++ b/spec/prometheus/middleware/collector_spec.rb
@@ -95,6 +95,20 @@ describe Prometheus::Middleware::Collector do
     expect(registry.get(metric).get(labels: labels)).to include("0.1" => 0, "0.5" => 1)
   end
 
+  it 'normalizes paths containing signed_ids by default' do
+    expect(Benchmark).to receive(:realtime).and_yield.and_return(0.3)
+
+    get '/foo/eyJfcmFpbHMiOnsibWVzc2FnZSI6--a419915169c2dc3419/bars'
+
+    metric = :http_server_requests_total
+    labels = { method: 'get', path: '/foo/:signed_id/bars', code: '200' }
+    expect(registry.get(metric).get(labels: labels)).to eql(1.0)
+
+    metric = :http_server_request_duration_seconds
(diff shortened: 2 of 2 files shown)


lib/prometheus/middleware/collector.rb
```diff
@@ -91,6 +91,7 @@ def strip_ids_from_path(path)
         path
           .gsub(%r{/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(/|$)}, '/:uuid\\1')
           .gsub(%r{/\d+(/|$)}, '/:id\\1')
+          .gsub(%r{/[0-9a-zA-Z=]+--[0-9a-f]+(/|$)}, '/:signed_id\\1')
```
dmagliola, line 94: Thank you for your PR!

I'm a little worried that this Regex is a bit broad. It'll match almost any part of a path that has `--` in the middle. This is admittedly rare, but i'm wondering if we can narrow it down a bit more, probably by specifying the length of these parts, like we do for `uuid`.

I have no experience with ActiveStorage unfortuantely, but reading through the code, i believe this is where the Signed ID gets generated: https://github.com/rails/rails/blob/afc79e3cb0906d06b035ef30411c44ddc2132409/activesupport/lib/active_support/message_verifier.rb#L186-L203

And if i'm reading that correctly, i'd expect the first half of the signed ID (before the `--`) can be any length, but the digest should always be the same length (40 chars), right?

I notice your test case has a shorter digest, i'm not sure whether you're using an actual Signed ID generated by Rails, which would invalidate my suggestion, or if it was manual random typing...

Thoughts?

rzjfr: Thank you for reviewing the PR,
I completely agree with that point, we should make the regex stricter.
Regarding the signature id, though my example was bogus, unfortunately lengths is not always 40, thought the default is SHA1, but it could be changed:
https://github.com/rails/rails/blob/afc79e3cb0906d06b035ef30411c44ddc2132409/activesupport/lib/active_support/message_verifier.rb#L109

Most likely to something like SHA-256 which is 64 chars but extremely rare (rarer than "having -- in path" rare) for anyone to change it to anything less than MD5 which is 32.

Also the first part (before --) also should be a valid base64 encoded string so a not a complex overkill  regex might look like this:
`(?=(.{4})*--)[A-Za-z0-9]+={0,2}--[0-9a-f]{32,}`

I can also add negative test case to the specs.

Sinjo:
Hi @rzjfr, sorry we went quiet on this PR.

After [some issues](https://github.com/prometheus/client_ruby/issues/249) caused by us trying to get too clever with how we process URLs in the collector middleware, we've decided to keep it simple for now.

For the time being, we're encouraging anyone who wants to customise how paths are generated to extend `Prometheus::Middleware::Collector` and [override one or both of `strip_ids_from_path` and `generate_path`](https://github.com/prometheus/client_ruby/blob/8740786a4b46f46c1c3ad893a699c3ce51d468b9/examples/rack/README.md#collector).

I appreciate that it's a little more code on your side, but we'd like to keep our path generation logic deliberately simple, and let people add their framework-specific customisations to it.

At some point we may revisit that decision, but the earliest that will happen is version 5 we've been treating changes to automatically generated labels as breaking (since they might break dashboards/alerts that people have set up).

PR #215, opened 2021-03-20
Minor wordsmithing in README.md

Sinjo:


README.md
```diff
@@ -389,10 +389,11 @@ If you are running in pre-fork servers (such as Unicorn or Puma with multiple pr
...
 files created by other processes on *this* run, instead of deleting old files.
 
-**Define metrics before fork**: In order to share the same definition of metrics between
-your forks, you should make sure to define your metrics before to fork your process.
-Indeed, even if the content of the metric is stored in a file, the list of all metrics is
-stored in memory. Creating metrics after fork would lead to unexported files metrics.
+**Declare metrics before fork**: As well as deleting files before your process forks, you
+should make sure to declare your metrics then too. Because the metric registry is held in
+memory, any metrics declared after forking will only be present in child processes where
+the code declaring them ran, and as a result won't be consistently exported when scraped
```
dmagliola, line 395: I'd say "may not be" rather than "won't". 
The reason is that if your code consistently declares metrics (and i'm not sure how this doesn't happen), you can totally declare them after fork and it's fine. For example, if your metrics get declared when Ruby executes the class definitions.

README.md
```diff
@@ -389,10 +389,11 @@ If you are running in pre-fork servers (such as Unicorn or Puma with multiple pr
 make sure you do this **before** the server forks. Otherwise, each child process may delete
 files created by other processes on *this* run, instead of deleting old files.
 
-**Define metrics before fork**: In order to share the same definition of metrics between
-your forks, you should make sure to define your metrics before to fork your process.
-Indeed, even if the content of the metric is stored in a file, the list of all metrics is
-stored in memory. Creating metrics after fork would lead to unexported files metrics.
+**Declare metrics before fork**: As well as deleting files before your process forks, you
+should make sure to declare your metrics then too. Because the metric registry is held in
```
dmagliola, line 393: `declare your metrics then too`
I find that "then too" a bit unclear as to when it is.

Can we make it more redundant but clear, something like `you should also make sure to declare your metrics before the fork`

dmagliola (approved):
2 nitpicky, optional comments. feel free to ignore if you disagree.

dmagliola (approved):
Love it, thank you

Issue #186, opened 2020-05-13
Improve path handling for Pushgateway


beorn7:
Currently, the code in `push.rb` only handles pushes to `/metrics/job/%s` and `'/metrics/job/%s/instance/%s`, and even those have some problems:

- If the job name or the instance name contain a `/`, the push will fail.
- If the instance name is empty, the push will fail. (It will also fail with an empty job name, but that's intended. This case could be caught earlier in the code, though.)
- No other grouping keys are possible.

Please check out https://github.com/prometheus/pushgateway/blob/master/README.md to learn about arbitrary label pairs as grouping key and how to handle base64 encoding for `/` and empty values (the latter to be merged as https://github.com/prometheus/pushgateway/pull/346 ).

Sinjo:
Hi @beorn7, I'm just running through our open issues and I wanted to ask your opinion on this one - particularly on handling special characters. I had a read through the `README` you linked and saw the option of encoding grouping key values as base64, which as far as I can tell deals with all possible encoding issues.

Would it be a reasonable approach to do that all the time rather than trying to detect if a particular value needs encoding? My rough plan, assuming that's a sensible approach, would be:

  - Add a way of specifying additional grouping keys when using `push.rb`
  - Validate the names of those keys using [the standard allowed characters for metric labels](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)
  - Suffix all keys with `@base64`
  - Encode all values as RFC4648 base 64

Alternatively, if that's overkill (I can see how it would make debugging more of a pain when using simple examples), I'm thinking:

  - Add a way of specifying additional grouping keys when using `push.rb`
  - Validate the names of those keys using [the standard allowed characters for metric labels](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)
  - If a value is empty or contains a `/`, suffix its key with `@base64` and encode the value as RFC4648 base 64
  - Otherwise apply standard URI encoding to the value

beorn7:
I guess you could just follow what I did in client_golang. Just always doing base64 is certainly a safe way to go, but I could see a loss of "quality of life" if you want to debug something, dump the URLs somehow, and then you have to decipher the base64 encoded parameters.

As far as I can see, past me made an effort to only use base64 encoding if required: https://github.com/prometheus/client_golang/blob/master/prometheus/push/push.go#L309-L319

Sinjo:
[...]

PR #220, opened 2021-03-27
Improve path handling in pushgateway client

Sinjo:
This implements some of the changes from #186. This is an intermediate PR to avoid building up too many unmerged commits.

### DIFF
--- a/lib/prometheus/client/push.rb
+++ b/lib/prometheus/client/push.rb
@@ -22,6 +22,10 @@ module Prometheus
       attr_reader :job, :instance, :gateway, :path
 
       def initialize(job, instance = nil, gateway = nil, **kwargs)
+        unless job
+          raise ArgumentError, "job cannot be nil"
+        end
+
         @mutex = Mutex.new
         @job = job
         @instance = instance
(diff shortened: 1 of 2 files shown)


lib/prometheus/client/push.rb
```diff
@@ -23,6 +23,7 @@ class Push
 
       def initialize(job:, instance: nil, gateway: DEFAULT_GATEWAY, **kwargs)
         raise ArgumentError, "job cannot be nil" if job.nil?
+        raise ArgumentError, "job cannot be empty" if job.empty?
```
dmagliola, line 26: Shall we do `!job || job.empty?` (you're effectively reimplementing `.blank?` here, i'd do literally the same thing, and in just one statement).
Opinion, ignore me if you disagree

Sinjo: Not mega fussed either way. For the sake of the extra line I'm gonna keep the more specific error messages.

Sinjo:
Merging this since you've already reviewed, and it's nice not to have any more commits build up on this branch! Will spin up another branch to finish the work for #186.

PR #222, opened 2021-04-05
Enable configuring `Net::HTTP` timeouts

nickvanw:
👋 !

We recently had an issue in our infrastructure where a Push Gateway was getting overloaded due to too much metric cardinality in the time series data that we were sending. We remedied this issue and things are well, but part of the way we noticed was that these pushes were happening in our job workers, which we have a lot of timing instrumentation around.

Digging in a bit further, we noticed that a huge number jobs were taking 60 seconds to complete, when they are normally much (sub-second) faster. We realized that this was due to our calls out to the push gateway, which would eventually time out. We wanted to be able to set the push timeout to be significantly lower (1-5 seconds) so that we can fail fast and avoid eating into our job capacity if there is ever an issue with our gateways again.

Looking at the [Ruby documentation for Net::HTTP](https://ruby-doc.org/stdlib-2.7.2/libdoc/net/http/rdoc/Net/HTTP.html), both `read_timeout` and `open_timeout` have a default timeout of 60 seconds:

```
If the HTTP object cannot read data in this many seconds, it raises a Net::ReadTimeout exception. The default value is 60 seconds.
```

```
If the HTTP object cannot open a connection in this many seconds, it raises a Net::OpenTimeout exception. The default value is 60 seconds.
```

This PR allows a user to pass through the `read_timeout` and `open_timeout` in the initialization of a new `Push` class, but retains the defaults otherwise. I am not in love with using `**kwargs`, and am open to other methods of configuring this value (and happy to implement them myself). I do believe that it's important to give the user control of these values however. 

Thanks!

### DIFF
--- a/lib/prometheus/client/push.rb
+++ b/lib/prometheus/client/push.rb
@@ -20,7 +20,7 @@ module Prometheus
 
       attr_reader :job, :instance, :gateway, :path
 
-      def initialize(job, instance = nil, gateway = nil)
+      def initialize(job, instance = nil, gateway = nil, **kwargs)
         @mutex = Mutex.new
         @job = job
         @instance = instance
@@ -30,6 +30,8 @@ module Prometheus
 
         @http = Net::HTTP.new(@uri.host, @uri.port)
         @http.use_ssl = (@uri.scheme == 'https')
+        @http.open_timeout = kwargs[:open_timeout] if kwargs[:open_timeout]
+        @http.read_timeout = kwargs[:read_timeout] if kwargs[:read_timeout]
       end
 
       def add(registry)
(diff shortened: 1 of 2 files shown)


Sinjo:
Hey @nickvanw, thanks for the PR!

I agree we should have a way to pass these through.

I'm actually in the middle of a rework of the pushgateway client over in #220 that switchers the initializer over to keyword arguments entirely (we missed updating this class to them in the big rework in #95) so I'm totally cool with using them.

The only change I'd ask for before merging is to add a test for these extra arguments. Just something [like this](https://github.com/prometheus/client_ruby/blob/a32180a2d779fb80cbf141f4666a1e4aa918d929/spec/prometheus/client/push_spec.rb#L123) that checks we're passing the parameters along.

nickvanw:
Thanks @Sinjo! 

I updated the fixture to take these arguments, and all of the relevant tests to check for them. Let me know if you want a separate test specifically for this (or, even, a specific instance of the gateway).

Sinjo (approved):
LGTM

I'm gonna refactor these tests so that the expectations aren't so duplicated (what with the `use_ssl` assertion already being repeated a bunch), but I'm not gonna hold up this PR with that.

Thanks for the contribution!

PR #224, opened 2021-04-19
implement and test our bugfix #223

### LINKED ISSUE #223: K8s nfs shared store file writing error
Hi,

We started to use your gem few weeks ago and I wanted to start by thanking you all about what you made.

We use this Gem in a context of a Rails 5.2 / sidekiq application that we run in separated pods all the components :

- Rails application in a pod, using [Yabeda::Prometheus::Exporter](https://github.com/yabeda-rb/yabeda-prometheus) to access the metrics on /metrics and the files are read from /app/tmp/metrics which is a [PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) with `ReadWriteMany` access mode.
- Sidekiq workers has access to this volume at the same path, and are in charge through an ActiveJob to collect the metrics every minutes (ActiveJob created by a dedicated cron pod)

We experienced rapidly and error (Errno::ESTALE) :

> /bundle/ruby/2.5.0/gems/prometheus-client-2.1.0/lib/prometheus/client/data_stores/direct_file_store.rb:284:in `seek'
> /bundle/ruby/2.5.0/gems/prometheus-client-2.1.0/lib/prometheus/client/data_stores/direct_file_store.rb:284:in `write_value'                                               
> /bundle/ruby/2.5.0/gems/prometheus-client-2.1.0/lib/prometheus/client/data_stores/direct_file_store.rb:104:in `block in set'                                              
> /bundle/ruby/2.5.0/gems/prometheus-client-2.1.0/lib/prometheus/client/data_stores/direct_file_store.rb:161:in `block in in_process_sync'                                  
> /usr/local/lib/ruby/2.5.0/monitor.rb:235:in `mon_synchronize'                                                                                                                                                      
> /bundle/ruby/2.5.0/gems/prometheus-client-2.1.0/lib/prometheus/client/data_stores/direct_file_store.rb:161:in `in_process_sync'
> ...

Since it is an error that may be raised by the filesystem, we don't see where to manage it elsewhere than in your library code.

So I'm about to create a PullRequest that should fix it with a test case.

I'll be grateful if you can challenge it in the form and content.

Best

ShamoX:
Proposition to fix #223 

Signed-off-by: Roland Laurès <roland@octopuslab.fr>

### DIFF
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -284,6 +284,16 @@ module Prometheus
             @f.seek(pos)
             @f.write([value, now].pack('dd'))
             @f.flush
+          rescue SystemCallError => error
+            # Question to Community: Maybe we should not warn here ? Or maybe we should.
+            warn "Catching SystemCallError while writing #{error}"
+            path = @f.path
+            # do not fail on close.
+            begin @f.close ensure @f = nil end
+            # re-open and all
+            initialize(path)
+            # retry and then fail.
+            write_value(key, value)
           end
 
           def close
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -274,7 +274,7 @@ module Prometheus
             @f.read(8).unpack('d')[0]
           end
 
-          def write_value(key, value)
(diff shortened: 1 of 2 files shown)


dmagliola:
When closing issue #223 , I forgot to also close this PR.
As explained in that issue, we're not planning to support this use case

Issue #225, opened 2021-05-12
`with_labels` doesn't seem to work


onlynone:
When I create a new `Counter` object from another by using the `Prometheus::Client::Metric#with_labels` method, it seems like the new `Counter` uses a different store from the original, so its metrics are never reported (say to the push client). And if I try to register the new `Counter` with the registry, I get an error because the original `Counter` is already registered with that name.

```ruby
#!/usr/bin/env ruby

require 'cgi'
require 'prometheus/client'
require 'prometheus/client/counter'
require 'prometheus/client/push'

counter1 = Prometheus::Client::Counter.new(
  :with_labels_counter,
  docstring: "with_labels counter",
  labels: [:foo]
)

counter1.increment(labels: {foo: "bar"})
counter1.increment(labels: {foo: "baz"})
counter1.increment(labels: {foo: "bat"})

counter2 = counter1.with_labels(foo: "bar")

counter2.increment

# I would expect this to print: [2.0, 1.0, 1.0]
# but it prints: [1.0, 1.0, 1.0]
puts "The counter values:"
puts ['bar', 'baz', 'bat'].map{|foo| counter1.get(labels: { foo: foo })}.inspect
puts ""

registry = Prometheus::Client.registry
registry.register(counter1)

push = Prometheus::Client::Push.new('with_labels_job', 'with_labels_instance')

push.add(registry)

resp = Net::HTTP.get_response(URI(push.gateway) + "metrics")

# I would expect the following to print:
#
# push_failure_time_seconds{instance="with_labels_instance",job="with_labels_job"} 0
# push_time_seconds{instance="with_labels_instance",job="with_labels_job"} 1.6208445893150244e+09
# # HELP with_labels_counter with_labels counter
# # TYPE with_labels_counter counter
# with_labels_counter{foo="bar",instance="with_labels_instance",job="with_labels_job"} 2
# with_labels_counter{foo="bat",instance="with_labels_instance",job="with_labels_job"} 1
# with_labels_counter{foo="baz",instance="with_labels_instance",job="with_labels_job"} 1
#
# But it prints:
#
# push_failure_time_seconds{instance="with_labels_instance",job="with_labels_job"} 0
# push_time_seconds{instance="with_labels_instance",job="with_labels_job"} 1.6208445893150244e+09
# # HELP with_labels_counter with_labels counter
# # TYPE with_labels_counter counter
# with_labels_counter{foo="bar",instance="with_labels_instance",job="with_labels_job"} 1
[...]

PR #227, opened 2021-06-12
Fix `with labels` bug reported in issue #225

dmagliola:
When calling `with_labels` on a metric object (let's call it "the original"), we instantiate a new metric (the "clone") that is identical except that it has some more pre-set labels, that allow the caller to observe it without having to specify the labels every time.

"currying", if you will.

The problem with the existing code (as exemplified by issue #225), is that as part of making this new metric, we end up instantiating a new store for this metric.

With in-memory stores, the new one will be empty. With file stores, it'll bring over the data from the original metric until the 
gets observed once, at which point they fork, while pointing at the same file and keeping separate internal state. An almost sure recipe for file corruption.

And when exporting, only the data in the "original" metric's store will be exported, the clone's will be ignored, assuming files didn't get corrupted.

The fix is not particularly elegant, but I don't see any way around it: we replace the internal store of the "clone" metric with the 
the "original", through the use of a protected method.

The only real alternative is getting rid of `with_labels`, which is a nice-to-have for convenience and performance, but not a necessity.

This PR is a "patch" version bump, and we should cherry-pick it out of master and cut a new patch version (as `master` contains breaking changes intended for `3.0` at the moment)

### DIFF
--- a/lib/prometheus/client/metric.rb
+++ b/lib/prometheus/client/metric.rb
@@ -41,8 +41,15 @@ module Prometheus
           metric_settings: store_settings
         )
 
+        # WARNING: Our internal store can be replaced later by `with_labels`
+        # Everything we do after this point needs to still work if @store gets replaced
         init_label_set({}) if labels.empty?
       end
+
+      protected def replace_internal_store(new_store)
+        @store = new_store
+      end
+
 
       # Returns the value for the given label set
       def get(labels: {})
@@ -51,11 +58,17 @@ module Prometheus
       end
 
       def with_labels(labels)
-        self.class.new(name,
-                       docstring: docstring,
-                       labels: @labels,
-                       preset_labels: preset_labels.merge(labels),
-                       store_settings: @store_settings)
+        new_metric = self.class.new(name,
+                                     docstring: docstring,
+                                     labels: @labels,
+                                     preset_labels: preset_labels.merge(labels),
+                                     store_settings: @store_settings)
+
+        # The new metric needs to use the same store as the "main" declared one, otherwise
+        # any observations on that copy with the pre-set labels won't actually be exported.
+        new_metric.replace_internal_store(@store)
+
+        new_metric
       end
 
       def init_label_set(labels)
--- a/spec/prometheus/client/counter_spec.rb
+++ b/spec/prometheus/client/counter_spec.rb
@@ -3,6 +3,7 @@
 require 'prometheus/client'
 require 'prometheus/client/counter'
 require 'examples/metric_example'
+require 'prometheus/client/data_stores/direct_file_store'
 
 describe Prometheus::Client::Counter do
   # Reset the data store
@@ -44,12 +45,6 @@ describe Prometheus::Client::Counter do
             counter.increment(labels: { test: 'label' })
           end.to change { counter.get(labels: { test: 'label' }) }.by(1.0)
         end.to_not change { counter.get(labels: { test: 'other' }) }
-      end
-
-      it 'can pre-set labels using `with_labels`' do
-        expect { counter.increment }
-          .to raise_error(Prometheus::Client::LabelSetValidator::InvalidLabelSetError)
-        expect { counter.with_labels(test: 'label').increment }.not_to raise_error
       end
     end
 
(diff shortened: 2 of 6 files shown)


spec/prometheus/client/counter_spec.rb
```diff
@@ -122,4 +117,57 @@
...
+    it 'registers `with_labels` observations in the original metric store' do
+      counter.increment(labels: { foo: 'value1'})
+      counter_with_labels = counter.with_labels({ foo: 'value2'})
+      counter_with_labels.increment(by: 2)
+
+      expect(counter_with_labels.values).to eql({foo: 'value1'} => 1.0, {foo: 'value2'} => 2.0)
+      expect(counter.values).to eql({foo: 'value1'} => 1.0, {foo: 'value2'} => 2.0)
+    end
+
+    context 'when using DirectFileStore' do
```
Sinjo, line 139: How come only `Counter` has specific tests for `DirectFileStore`?

As an aside: this PR is even more encouragement for the test refactor I wanted to do. These tests definitely fall into the category that I'd prefer to test via a subclass of `Metric` rather than separately per type of metric.

lib/prometheus/client/metric.rb
```diff
@@ -41,21 +41,37 @@ def initialize(name,
           metric_settings: store_settings
         )
 
+        # WARNING: Our internal store can be replaced later by `with_labels`
+        # Everything we do after this point needs to still work if @store gets replaced
         init_label_set({}) if labels.empty?
       end
 
+      protected
```
Sinjo, line 49: If you're only making one method `protected`, you can put it in front of `def` to avoid having to switch back to `public` after.

spec/prometheus/client/counter_spec.rb
```diff
@@ -122,4 +117,57 @@
...
+        .to raise_error(Prometheus::Client::LabelSetValidator::InvalidLabelSetError)
+      expect { counter.with_labels(foo: 'label').increment }.not_to raise_error
+    end
+
+    it 'registers `with_labels` observations in the original metric store' do
+      counter.increment(labels: { foo: 'value1'})
+      counter_with_labels = counter.with_labels({ foo: 'value2'})
+      counter_with_labels.increment(by: 2)
+
+      expect(counter_with_labels.values).to eql({foo: 'value1'} => 1.0, {foo: 'value2'} => 2.0)
```
Sinjo, line 135: Could compare the two `values` with each other rather than looking for specific values in the hash. You're already testing that the thing returned by `with_labels` works above.

spec/prometheus/client/counter_spec.rb
```diff
@@ -122,4 +117,57 @@
...
+        Dir.glob('/tmp/prometheus_test/*').each { |file| File.delete(file) }
+        Prometheus::Client.config.data_store = Prometheus::Client::DataStores::DirectFileStore.new(dir: '/tmp/prometheus_test')
+      end
+
+      let(:expected_labels) { [:foo, :bar] }
+      
+      it "doesn't corrupt the data files" do
+        counter_with_labels = counter.with_labels({ foo: 'longervalue'})
+
+        # Initialize / read the files for both views of the metric
```
Sinjo, line 150: The steps in this test feel very particularly chosen, but it's not clear to a reader why. Would you be up for writing up why these operations specifically would cause corruption in a comment?

Sinjo (commented):
Few comments, mostly looks good!

spec/prometheus/client/counter_spec.rb
dmagliola: For the very reason you mentioned. I just need to test it in *one*, and `Counter` is kinda the default. We can definitely test it through a subclass of `Metric`.
I can also copy it to the other 3 metrics if you want, but it seemed a bit pointless

lib/prometheus/client/metric.rb
dmagliola: Yes... A feature they added in *some* version of Ruby, but I can never find which one, because I don't know what it's called... So i'm not sure we can do that in all the versions we support. If you are sure, I can totally do that.

spec/prometheus/client/counter_spec.rb
dmagliola: you mean expect counter_with_labels.values? == counter.values?

spec/prometheus/client/counter_spec.rb
dmagliola: No. I prefer people to work it out for themselves.

dmagliola: Just kidding,that's a good idea, i'll do that tomorrow :)

spec/prometheus/client/counter_spec.rb
dmagliola: If you do mean that, i prefer this way, because otherwise those hashes could be empty for all i know...

lib/prometheus/client/metric.rb
dmagliola: Apparently it's from 2.1.
So, yeah, fair enough, 2.0 is pretty old by now

spec/prometheus/client/counter_spec.rb
dmagliola: Added a 3rd commit with the explanation. It's *long*, sorry, I don't know how to properly explain it shorter without the explanation being more cryptic than the code.

spec/prometheus/client/counter_spec.rb
Sinjo: Yeah, that is what I mean.

There are other tests that assert that labels behave in a reasonable way in general. I wouldn't expect them to suddenly do something wierd (i.e. be completely wiped out in both the original and new counters) by a call to `with_labels`.

As a testing style thing, I try not to have too many tests checking the same property of the code over and over again.

spec/prometheus/client/counter_spec.rb
Sinjo: Yeah, no point copying it multiple times. Let's have it in `Counter` for now, and then I'll go refactor the fuck out of these tests in a separate PR.

spec/prometheus/client/counter_spec.rb
dmagliola: How strongly do you feel about this?
Because I think the test, as is, is more dependable (as it's more obvious what the exact contents of the hash is)
But if that's going to block merging this fix, i'll just change it.

spec/prometheus/client/counter_spec.rb
Sinjo: Resolving this. I'll get it in that refactor of the metric tests.

spec/prometheus/client/counter_spec.rb
Sinjo: Really not that strong. Definitely not strongly enough to stop this fix going out. Resolving.

dmagliola: Shall we merge it in then?

Sinjo (approved):

PR #234, opened 2021-11-21
Finish improvements to push client

Sinjo:
This PR implements the remaining improvements from #186 that weren't in #220.

### DIFF
--- a/lib/prometheus/client/push.rb
+++ b/lib/prometheus/client/push.rb
@@ -5,6 +5,7 @@ require 'thread'
 require 'net/http'
 require 'uri'
 require 'erb'
+require 'set'
 
 require 'prometheus/client'
 require 'prometheus/client/formats/text'
@@ -103,6 +104,8 @@ module Prometheus
       end
 
       def request(req_class, registry = nil)
+        validate_no_label_clashes!(registry) if registry
+
         req = req_class.new(@uri)
         req.content_type = Formats::Text::CONTENT_TYPE
         req.basic_auth(@uri.user, @uri.password) if @uri.user
@@ -113,6 +116,25 @@ module Prometheus
 
       def synchronize
         @mutex.synchronize { yield }
+      end
+
+      def validate_no_label_clashes!(registry)
+        # There's nothing to check if we don't have a grouping key
+        return if @grouping_key.empty?
+
+        # We could be doing a lot of comparisons, so let's do them against a
+        # set rather than an array
+        grouping_key_labels = @grouping_key.keys.to_set
+
+        registry.metrics.each do |metric|
+          metric.labels.each do |label|
+            if grouping_key_labels.include?(label)
+              raise LabelSetValidator::InvalidLabelSetError,
+                "label :#{label} from grouping key collides with label of the " \
+                "same name from metric :#{metric.name} and would overwrite it"
+            end
+          end
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


Sinjo:
I'll follow up with changes to `README.md` once these changes are approved!

spec/prometheus/client/push_spec.rb
```diff
@@ -111,12 +142,99 @@
...
       expect(http).to receive(:open_timeout=).with(5)
       expect(http).to receive(:read_timeout=).with(30)
-      expect(http).to receive(:request).with(request)
+      expect(http).to receive(:request).with(request).and_return(response)
       expect(Net::HTTP).to receive(:new).with('localhost', 9091).and_return(http)
 
       push.send(:request, Net::HTTP::Post, registry)
     end
 
+    context 'for a 3xx response' do
```
Sinjo, line 151: I was on the fence about making shared examples for these. On the one hand there's 3 very similar tests that can be parameterised. On the other, there are exactly 3 cases and that won't grow.

Overall I'm not a huge fan of these tests. One thought I had to reduce the level of repetition is to pull `request` and `http` into `let` blocks with all the appropriate method stubs set up. After that, each test could set the up the method call expectations that it specifically cares about (e.g. you'd only set up the `use_ssl` expectation in the test for SSL).

If you think that'd be better I'm happy to follow up either with another commit on this branch, or with a separate refactor PR.

lib/prometheus/client/push.rb
```diff
@@ -70,26 +79,85 @@ def parse(url)
...
+          # For empty strings, we use a single padding character (`=`) as the
+          # value.
+          #
+          # See the pushgateway docs for more details:
+          #
+          # https://github.com/prometheus/pushgateway/blob/6393a901f56d4dda62cd0f6ab1f1f07c495b6354/README.md#url
+          elsif value.empty?
+            path += "/#{label}@base64/="
+          else
+            path += "/#{label}/#{value}"
```
Sinjo, line 104: In case you review this before I fix it: just realised this needs `ERB::Util::url_encode` applying to it too.

spec/prometheus/client/push_spec.rb
dmagliola: I think this is fine?
i'm not a big fan of "shared examples", they tend to make things harder to both read and modify.
Agree you can extract those two very mocked objects to centralize them, if you feel strongly about it. Otherwise, I'd leave this as is until one of those blocks needs to be modified 3 times, and then i'd extract.

dmagliola (commented):
Nice set of improvements!! ❤️

dmagliola (approved):
Sorry, meant to approve in the previous review!

lib/prometheus/client/push.rb
```diff
@@ -68,8 +69,14 @@ def parse(url)
         raise ArgumentError, "#{url} is not a valid URL: #{e}"
       end
 
-      def build_path(job)
-        format(PATH, ERB::Util::url_encode(job))
+      def build_path(job, grouping_key)
+        path = format(PATH, ERB::Util::url_encode(job))
+
+        grouping_key.each_pair do |label, value|
```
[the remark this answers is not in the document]
Sinjo: Nothing more than habit. Will change it.

spec/prometheus/client/push_spec.rb
Sinjo: Yeah, shared examples specifically I'm not a huge fan of either. I think I'm going to go the route of extracting as much of the duplication as possible and not setting up the same mock expectations in every test.

I'll do that in a separate PR though, so we can get this one merged.

PR #230, opened 2021-06-20
Convert to CircleCI for build testing

SuperQ:
Replace Travis with CircleCI for testing.

Signed-off-by: SuperQ <superq@gmail.com>

PR #239, opened 2022-01-07
Replace Coveralls with SimpleCov

dmagliola:
We've been having errors in CI from Coveralls, when trying to upload results.
(As mentioned in [this comment](https://github.com/prometheus/client_ruby/pull/230#issuecomment-864561231))

This error seems to be because our coveralls gem is pretty old and abandoned,
and probably using a TLS version that is no longer supported.

(Reference: https://github.com/lemurheavy/coveralls-ruby/issues/163)

One option recommended in that issue is to switch to a different `coveralls-ruby-reborn` gem.

However, given that we only use `coveralls` to upload results to the cloud,
only so we can have a badge in our README reporting 100%, in the interest of
security, I think i'd rather get rid of `coveralls` altogether, and use
`simplecov` directly instead, which reports the coverage when running the
tests and doesn't upload them anywhere.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :test do
-  gem 'coveralls'
+  gem 'simplecov'
   gem 'json'
   gem 'rack'
   gem 'rack-test'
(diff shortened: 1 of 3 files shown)


dmagliola:
@Sinjo @SuperQ what do you think?

Should we move to [`coveralls-ruby-reborn`](https://github.com/tagliala/coveralls-ruby-reborn), which I don't know much about?
Or should we get rid of that badge and stop using `coveralls` entirely?

Is there any other use than that badge for coveralls.io?

Sinjo (approved):
I'm +1 on this. I don't think the README badge is worth fighting poorly-maintained Ruby gems.

It might be nice for us to do something more with the report, but we can do that elsewhere.

PR #242, opened 2022-01-09
Move to explicit method call for Basic Auth credentials in push client

### LINKED ISSUE #170: Pushgateway basic auth with special characters in password
Looks like we can't to make basic auth with special characters in the password. I've implemented fast monkey-patching
```
module PrometheusClientPush
  attr_reader :basic_auth

  def set_basic_auth(username, password)
    @basic_auth = OpenStruct.new(username: username, password: password)
  end

  private

  def request(req_class, registry = nil)
    req = req_class.new(@uri)
    req.content_type = Prometheus::Client::Formats::Text::CONTENT_TYPE
    req.basic_auth(@uri.user || basic_auth.username, @uri.password || basic_auth.password) if @uri.user || basic_auth
    req.body = Prometheus::Client::Formats::Text.marshal(registry) if registry

    @http.request(req)
  end
end

Prometheus::Client::Push.prepend(PrometheusClientPush)
```

Will be great to see this possibility in a new version of gem, perhaps with better implementation

Sinjo:
While URLs do support passing Basic Auth credentials using the
`http://user:password@example.com/` syntax, the username and password in
that syntax have to follow the usual rules for URL encoding of
characters per RFC 3986
(https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).

Rather than place the burden of correctly performing that encoding on
users of this gem, we decided to have a separate method for supplying
Basic Auth credentials, with no requirement to URL encode the characters
in them.

---

Since the tests around HTTP requests are extremely heavily mocked, I tested this locally with a real pushgateway configured with one user with `@` in the password and another with `:` in it, as well as a more boring password that only used basic latin alphabet characters. All three worked.

Maybe at some point we should have some integration tests, but that feels like its own piece of work.

Fixes #170

### DIFF
--- a/lib/prometheus/client/push.rb
+++ b/lib/prometheus/client/push.rb
@@ -39,12 +39,19 @@ module Prometheus
         @gateway = gateway || DEFAULT_GATEWAY
         @grouping_key = grouping_key
         @path = build_path(job, grouping_key)
+
         @uri = parse("#{@gateway}#{@path}")
+        validate_no_basic_auth!(@uri)
 
         @http = Net::HTTP.new(@uri.host, @uri.port)
         @http.use_ssl = (@uri.scheme == 'https')
         @http.open_timeout = kwargs[:open_timeout] if kwargs[:open_timeout]
         @http.read_timeout = kwargs[:read_timeout] if kwargs[:read_timeout]
+      end
+
+      def basic_auth(user, password)
+        @user = user
+        @password = password
       end
 
       def add(registry)
@@ -113,7 +120,7 @@ module Prometheus
 
         req = req_class.new(@uri)
         req.content_type = Formats::Text::CONTENT_TYPE
-        req.basic_auth(@uri.user, @uri.password) if @uri.user
+        req.basic_auth(@user, @password) if @user
         req.body = Formats::Text.marshal(registry) if registry
 
         response = @http.request(req)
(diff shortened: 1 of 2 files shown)


lib/prometheus/client/push.rb
```diff
@@ -39,14 +39,21 @@ def initialize(job:, gateway: DEFAULT_GATEWAY, grouping_key: {}, **kwargs)
...
         @uri = parse("#{@gateway}#{@path}")
+        validate_no_basic_auth!(@uri)
 
         @http = Net::HTTP.new(@uri.host, @uri.port)
         @http.use_ssl = (@uri.scheme == 'https')
         @http.open_timeout = kwargs[:open_timeout] if kwargs[:open_timeout]
         @http.read_timeout = kwargs[:read_timeout] if kwargs[:read_timeout]
       end
 
+      def basic_auth(user, password)
```
Sinjo, line 52: I went with the same method name as `Net::HTTP` as it seems like a reasonable one, though I'd also be open to something like `set_basic_auth` to make it clearer if you'd prefer.

lib/prometheus/client/push.rb
```diff
@@ -126,6 +133,27 @@ def synchronize
         @mutex.synchronize { yield }
       end
 
+      def validate_no_basic_auth!(uri)
+        if uri.user || uri.password
+          raise ArgumentError, <<~EOF
+            Setting Basic Auth credentials in the gateway URL is not supported, please call the `basic_auth` method
```
nickcampbell18, line 139: Driveby 😄 maybe a quick usage example here that a developer could just copy-paste?

dmagliola (approved):
Beautiful error message

lib/prometheus/client/push.rb
Sinjo: That's a really good idea. Just added it in!

Also hi Nick 👋🏻

PR #228, opened 2021-06-19
Update CHANGELOG in preparation for 2.2.0 release

dmagliola:
Includes the changes in `master` that we'll release in `2.2.0`, and
the ones that we can't because of breaking changes, which we'll release
in `3.0.0`

--------

I'm proposing to release a new version with the bug fixes and backwards-compatible improvements we've introduced (which we should have released ages ago, but life), including the not-yet-merged fix to `with_labels` in #227.

Since `master` has many non-releasable changes, I've created branch `v2.2_release`, which forks off of `2.1.0` and has the relevant commits cherry-picked into it.

[Comparison between 2.2.0 (cherry-picked commits) vs 2.1.0](https://github.com/prometheus/client_ruby/compare/v2.1.0...v2.2_release?expand=1)

Once we finish the changes to PushGateway, we'll release v3.0 with what's in `master` right now.

--------

NOTE: Do not merge this yet. If @Sinjo agrees with this approach, I'll update this README with the correct, date, merge, and add a new release tag in the `v2.2_release` branch

PR #244, opened 2022-01-28
Update CHANGELOG.md and gem version for 3.0.0 release

### LINKED ISSUE #241: Update CHANGELOG.md for 3.0.0 release
As the title says.

Sinjo:
@dmagliola I've pulled in your work from #228, which we were going to use to cut an interim 2.2.0 release, and combined it all into one big section for the 3.0.0 release which we decided to go with instead.

Fixes #241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,13 @@ adapt to, but please read the details below:
 Please refer to [UPGRADING.md](UPGRADING.md) for details on upgrading from versions
 `< 3.0.0`.
 
-- [#206](https://github.com/prometheus/client_ruby/pull/206) Include `SCRIPT_NAME` when 
-    determining path in Collector:  
+- [#206](https://github.com/prometheus/client_ruby/pull/206) Include `SCRIPT_NAME` when
+    determining path in Collector:
     When determining the path for a request, `Rack::Request` prefixes the
     `SCRIPT_NAME`. This was a problem with our code when using mountable engines,
     where the engine part of the path gets lost. This patch fixes that to include `SCRIPT_NAME` as part of the path.
-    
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,122 @@
 # CHANGELOG
 
+# Unreleased changes
+
+_None outstanding_
```
dmagliola, line 5: Do we need this top section?

dmagliola (approved):
Nice :)
Looking forward to releasing this one!

CHANGELOG.md
Sinjo: We don't right now. I was putting it there as a prompt for us to build this up incrementally between future releases.

Do you dislike it being there?

Sinjo:
@dmagliola One question before we merge this: do you think it's worth us putting a release candidate out given the number of changes?

I'm thinking it might be worth it if we also post it on prometheus-announce so that there's more chance of people giving it a go. Equally, I know how reticent people are to use software that's labelled as a release candidate (but would be totally fine with using if it were a proper release), which limits their value.

dmagliola:
I've taken a look at the changes which we introduced, and I feel like we can be fairly confident that they'll work.

I'd put out a release candidate if we ever merge lzap's change to the DirectFileStore, that kind of change seems more like something that could have hard-to-predict implications.

What we've done in these commits is the kind of thing that is pretty testable, and it's tested. I think we can release as 3.0.0, to be honest.

dmagliola (approved):

CHANGELOG.md
dmagliola: 👍

PR #245, opened 2022-02-01
Use framework-specific route info and handle consecutive path segments containing IDs in Collector

### LINKED ISSUE #233: better path handling
I'm trying to use the collector rack middleware with a sinatra app

I have a simple endpoint `/add/:num_a/:num_b` that takes num_a and num_b and returns their sum

The  path label value generated by the collector middleware for that route is `path="/add/:id/5` , `path="/add/:id/3` ...

I think a batter solution would be to take the matching route pattern and use it as the path value, in my case, I'd expect to have `path="/add/:num_a/:num_b`

any thoughts about that?

Sinjo:
Fixes #233

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -140,7 +140,7 @@ recommends exporting a default value for any time series you know will exist in 
 For series with no labels, other Prometheus clients (including Go, Java, and Python) do
 this automatically, so we have matched that behaviour in the 3.x.x series.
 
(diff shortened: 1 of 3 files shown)


Sinjo:
@dmagliola I've added the changes to `UPGRADING.md` to this PR as well, to save going through another one.

I'm going to add the `CHANGELOG.md` entry to #244.

dmagliola (approved):

Issue #249, opened 2022-02-11
`generate_path` yields incorrect values for Sinatra+Rack::Build apps


shurikp:
`Prometheus::Middleware::Collector#generate_path` yields incorrect values when used on top-level of application build with `Rack::Build`, that also has several mounted Sinatra apps inside `map '/subpath'` blocks.

We have the following setup:

config.ru:
```
app = Rack::Builder.app do # rubocop:disable Metrics/BlockLength
  use Prometheus::Middleware::Exporter
  use Prometheus::Middleware::Collector

  map '/app1' do
    run App1
  end

  map '/app2' do
    run App2
  end
end

run app
```

app1.rb:
```
class App1 < Sinatra::Base
  get '/' do
     ...
  end
end
```

app2.rb:
```
class App2 < Sinatra::Base
  get '/' do
     ...
  end
end
```

After upgrading `prometheus-client` to 3.0.0, metrics start to go into wrong buckets: for both `/app1` and `/app2` requests we now have:
```http_server_request_duration_seconds_bucket{method="get",path="/",le="0.025"} 0.0```

path labels being "/" for both, whereas before it was:

```
http_server_request_duration_seconds_bucket{method="get",path="/app1",le="0.025"} 0.0
http_server_request_duration_seconds_bucket{method="get",path="/app2",le="0.025"} 0.0
```

The problem here is with the new way of inferring paths for sinatra applications: https://github.com/prometheus/client_ruby/pull/245/files#diff-1026444c3a547bf72b02d69fa3ca9be012988d69f45af85bbee6df35560c9dbfR108-R109
The problem is that `env['sinatra.route']` is present but `env['SCRIPT_NAME']` at the moment this top-level middleware runs (before we hit Sinatra app).

If we move `use Prometheus::Middleware::Collector` inside each of the applications they'll start to conflict with `http_server_requests_total has already been registered` error.

Current workaround is to just monkeypatch `generate_path` to the old behaviour:
```
def generate_path(env)
  [env['SCRIPT_NAME'], env['PATH_INFO']].join
end
```
but I wonder if there's any better solution

**Comment by Sinjo** [comment]
Okay, I did a little digging, and it's not quite what I expected. It looks like `Rack::Builder` makes modifications to `SCRIPT_NAME` and `PATH_INFO` when it routes a request into an app that you specified in a `map` block.

I printed those two values in two places: once in the Prometheus collector middleware, and once in the app itself. In the collector middleware, they look like this:

```
[...]

PR #251, opened 2022-03-08
Remove framework-specific route detection from collector middleware

Sinjo:
Sadly, with the benefit of hindsight, this wasn't a good idea.

There are two reasons we're dropping this:

  - It doesn't play nicely with libraries like `Rack::Builder`, which
    dispatches requests to different Rack apps based on a path prefix in
    a way that isn't visible to middleware.

    For example, when using `Rack::Builder`, `sinatra.route` only
    contains the parts of the path after the prefix that `Rack::Builder`
    used to dispatch to the specific app, and doesn't leave any
    information in the request environment to indicate which prefix it
    dispatched to.
  - It turns out framework-specific route info isn't always formatted in
    a way that looks good in a Prometheus label.

    For example, when using regex-based route-matching in Sinatra, you
    can end up with labels that look like `\\/vapes\\/([0-9]+)\\/?`.

For a really detailed dive into those two issues, see this GitHub
comment:

https://github.com/prometheus/client_ruby/issues/249#issuecomment-1061317511


Sinjo:
Sure thing. Maybe a slight rearrange of [this doc](https://github.com/prometheus/client_ruby/blob/134c1fd644d4ad2135a6412cb2035dd57fa90c4e/examples/rack/README.md#collector) to emphasise that you can customise `path` in its entirety?

dmagliola:
Yeah, that sounds perfect.

Sinjo:
All done!

dmagliola (approved):
Awesome, thank you

PR #254, opened 2022-04-11
Add optional preset_labels to Middleware::Collector metrics

JoshInLisbon:
Signed-off-by: Joshua Eichler-Summers <joshua.summers@boardintelligence.com>

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -14,6 +14,9 @@ module Prometheus
     # By default metrics all have the prefix "http_server". Set
     # `:metrics_prefix` to something else if you like.
     #
+    # By default metrics have no preset_labels. Set `:preset_labels` as a hash
+    # containing `label_name: 'value'` pairs to include them.
+    #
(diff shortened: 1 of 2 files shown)


dmagliola:
Hello @JoshInLisbon,

Unless i'm misunderstanding the intention here, this looks like a re-do of #180 and #174.

Generally speaking, "the Prometheus way" is to [set target labels on the Prometheus end](https://github.com/prometheus/client_ruby/pull/180#issuecomment-582947801), instead of reporting those labels from your app.

[Tobias explains it well here](https://github.com/prometheus/client_ruby/pull/93#issuecomment-429309390)

Based on this, I'm closing this PR as it goes against Prometheus Best Practices. Please reopen if I've misunderstood the intention here.

hosamaly:
I think that it's quite restrictive to not accept preset labels. There is a reason why they exist in the API. Some values may not be available to Prometheus (e.g. an app's version). There may be other legitimate reasons to not use target labels, such as when a team doesn't have access to the Prometheus configuration.

PR #260, opened 2022-06-17
Update readme

Juanmcuello:
This MR contains two commits with minor changes to the README.md.

The first one fixes a typo, removing a duplicated word.

The second one removes trailing whitespaces. Not sure if you are open to this, but I personally think it is an improvement. We can drop it if you don't want it and keep only the first commit.


Sinjo:
@Juanmcuello I'm +1 on both of these commits.

Would you be able to amend them with a signoff so that DCO check passes? More details on it [here](https://github.com/apps/dco).

Juanmcuello:
Done!

Sinjo (approved):

PR #261, opened 2022-06-24
Document startup file cleanup

Ehrhardt1979:
Spelling Readme


dmagliola:
Hi, could you please Rebase this against the main branch, so the PR is a bit cleaner?
I'm not entirely sure what we're looking at here

Sinjo:
Hi @Ehrhardt1979, I'm closing this for now due to the lack of reply. It's not clear what's being changed and the commit history really needs a cleanup, but if you resubmit after a cleanup we'd be happy to take a look.

PR #264, opened 2022-08-21
Add JRuby 9.3 to the build matrix

Sinjo:
I don't really follow JRuby any more, so this release passed me by, but
I figure it's worth adding to our build matrix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,3 +30,4 @@ workflows:
             - cimg/ruby:3.1
             - circleci/jruby:9.1
             - circleci/jruby:9.2
+            - circleci/jruby:9.3


dmagliola (approved):
On that note, Ruby 2.6 is EOL.
Should we remove it since we only officially support current versions?
Or should we leave it since it doesn't really cost us anything?

Sinjo:
I was tempted to remove it until I read this line in our comptibility docs:

> Whenever a version of Ruby falls out of support we will mirror that change in the Prometheus Ruby client by updating the build matrix and releasing a new major version.

I'm normally not too averse to bumping major version, but seeing as the carrying cost of having 2.6 in our CI matrix is basically 0, I'm happy to leave it there for now. It would feel slightly strange to bump to 5.0 without having released any features/improvements since 4.0.

PR #266, opened 2022-09-12
Allow Interpolated symbols for metric name

ddotafonso:
### Overview

The current implementation doesn't allow the use of interpolated symbols. On a use case where I'd like to dynamically generate different metric names via Ruby symbol interpolating i.e: `:"test_#{test_var}"`, is currently not possible, therefore, proposing this change.

For instance, ideally the following symbol definition should be welcomed as a metric name `:"metric_name_for_#{service_name}"` or even `"metric_name_for_#{service_name}".to_sym`, however, it returns ``prometheus-client-2.1.0/lib/prometheus/client/metric.rb:80:in `validate_name': metric name must match /[a-zA-Z_:][a-zA-Z0-9_:]*/ (ArgumentError)``

### DIFF
--- a/lib/prometheus/client/metric.rb
+++ b/lib/prometheus/client/metric.rb
@@ -90,8 +90,8 @@ module Prometheus
         unless name.is_a?(Symbol)
           raise ArgumentError, 'metric name must be a symbol'
         end
-        unless name.to_s =~ /\A[a-zA-Z_:][a-zA-Z0-9_:]*\Z/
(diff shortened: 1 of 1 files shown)


Sinjo:
Hi @ddotafonso,

The problem here isn't interpolation. The problem is that whatever variable you're interpolating into the symbol contains characters that aren't valid in a Prometheus metric name.

This isn't something we can change here, as it's a restriction [within the whole Prometheus ecosystem](https://prometheus.io/docs/concepts/data_model/).

PR #268, opened 2022-11-24
Use lowercase response headers in Rack example

Sinjo:
Because HTTP/2 requires that headers are sent in lowercase, Rack has started validating that no header contains an uppercase letter.

We were setting uppercase headers as that was a common convention in HTTP/1.1, where they were interpreted in a case-insensitive fashion.

This change makes us compatible with Rack 3.0 without breaking compatibiility for older versions (not that it would matter for this example code).

---

Fixes #267

PR #271, opened 2022-12-20
Using lowercase for HTTP headers to make it compatible with HTTP/2 and Rack v3

bitmaybewise:
Given Rack is validating HTTP headers has no uppercased letters, as mentioned in this previous PR https://github.com/prometheus/client_ruby/pull/268, the exporter needs to use lowercased letters to keep compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ use Rack::Deflater
 use Prometheus::Middleware::Collector
 use Prometheus::Middleware::Exporter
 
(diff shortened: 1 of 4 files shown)


Sinjo (approved):
Thanks for sending this in!

Out of an unreasonable level of paranoia, I've just done a quick check that Promethus still scrapes our example app with these changes, and everything is fine.

PR #272, opened 2022-12-25
Fix builds for JRuby <=9.2 in CircleCI

Sinjo:
A new version of `rubygems-update` was just released, which drops support for Ruby versions below 2.6.0. These older versions of JRuby target Ruby language versions earlier than that, so have been failing their CI runs.

This should get us back to green builds, and will be easy to remove once we drop support for the JRuby versions involved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,18 @@ jobs:
 
     steps:
     - checkout
-    - run: if [[ "$(ruby -e 'puts RUBY_VERSION')" != 1.* ]]; then gem update --system; fi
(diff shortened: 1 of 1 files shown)


dmagliola:
I'm confused by this... Shouldn't we remove unsupported versions from the jruby matrix instead?

Sinjo:
I believe 9.2 is still supported, so a version cull wouldn't get rid of the conditional for now. At some point in the future that's what we'll do.

dmagliola (approved):
Makes sense

PR #282, opened 2023-03-20
Fix release date in CHANGELOG.md

Sinjo:


Sinjo:
Given how utterly trivial this change was, I merged without review so I can actually cut the relase.

PR #285, opened 2023-04-02
Use Rack::Response#headers in place of header

### LINKED ISSUE #278: Fix `Rack` deprecation warning
Right now, our tests are spamming this:

```
./home/circleci/project/spec/prometheus/middleware/exporter_spec.rb:38: warning: Rack::Response#header is deprecated and will be removed in Rack 3.1
```

We should switch to using whatever the replacement is. From a quick grep, it looks like we only call that method in tests, so we shouldn't have to worry about maintaining compatibility with multiple versions in the library code.

```
$ git grep -n '\.header'
spec/prometheus/middleware/exporter_spec.rb:38:        expect(last_response.header['content-type']).to eql(fmt::CONTENT_TYPE)
spec/prometheus/middleware/exporter_spec.rb:50:        expect(last_response.header['content-type']).to eql('text/plain')
spec/prometheus/middleware/exporter_spec.rb:111:          expect(last_response.header['content-type']).to eql(text::CONTENT_TYPE)
```

duffn:
Use `headers` in place of deprecated `header` in tests per https://www.rubydoc.info/gems/rack/Rack%2FResponse:header

Closes #278

### DIFF
--- a/spec/prometheus/middleware/exporter_spec.rb
+++ b/spec/prometheus/middleware/exporter_spec.rb
@@ -35,7 +35,7 @@ describe Prometheus::Middleware::Exporter do
         get '/metrics', nil, headers
 
         expect(last_response.status).to eql(200)
-        expect(last_response.header['content-type']).to eql(fmt::CONTENT_TYPE)
+        expect(last_response.headers['content-type']).to eql(fmt::CONTENT_TYPE)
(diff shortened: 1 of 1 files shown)


Sinjo (approved):
Thank you for this!

PR #287, opened 2023-06-09
Add `Gauge#set_to_current_time`

### LINKED ISSUE #51: Support metric timestamps
Hi,

I'm using client_ruby to create/set counter and gauge metrics but i don't see in the prometheus library an option to specify a timestamp.

Is this not possibly?

Sinjo:
I was porting some code from Python to Ruby today, and when I went to call this method, I realised we don't have it!

My use-case is to track the last time some code successfully ran.

I'm open to discussing the naming, but I've defaulted to the name used in `client_python` because it seems good to me.

---

Fixes #51 (I think?)

### DIFF
--- a/lib/prometheus/client/gauge.rb
+++ b/lib/prometheus/client/gauge.rb
@@ -20,6 +20,10 @@ module Prometheus
         @store.set(labels: label_set_for(labels), val: value)
       end
 
+      def set_to_current_time(labels: {})
+        @store.set(labels: label_set_for(labels), val: Time.now.to_f)
+      end
+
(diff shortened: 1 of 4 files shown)


electron0zero (approved):

electron0zero:
I think we are missing it in client_ruby, and should add it, guide on writing client says Guage SHOULD have this method:

> A gauge SHOULD have the following methods:
> set_to_current_time(): Set the gauge to the current unixtime in seconds.

source: https://prometheus.io/docs/instrumenting/writing_clientlibs/#gauge

Sinjo:
Perfect. Let's do this!

Issue #293, opened 2023-08-30
cannot load such file -- prometheus/middleware/collector (LoadError)


DerekCrosson:
When running the config.ru code in the readme I get the following error:

cannot load such file -- prometheus/middleware/collector (LoadError)

This happens because the required files and namespaces have changed.|

This PR updates the readme to provide code that works: https://github.com/prometheus/client_ruby/pull/294

Sinjo:
Closing this and the assosiated PR as I'm pretty sure those paths are from an older version. Feel free to reopen if there's an issue with the current version.

PR #294, opened 2023-08-30
Update Readme - config.ru

DerekCrosson:
See: https://github.com/prometheus/client_ruby/issues/293


Sinjo:
See [here](https://github.com/prometheus/client_ruby/issues/293#issuecomment-1782468575) for reason this was closed.

PR #301, opened 2023-12-25
Add Ruby 3.3

Sinjo:
Merry Christmas! 🎄☃️

---

No image from CircleCI just yet, but I'll re-run the tests when it appears. The JRuby 9.3 failure looks like drift and I'll fix it in a separate PR.


dmagliola (approved):
Merry xmas 🍾

electron0zero (approved):
happy new year, and I see that Circle CI pushed 3.3 image on Jan 2, 2024.

I think we are good to merge this 🎊

PR #302, opened 2023-12-26
Fix JRuby 9.3 build

Sinjo:
The Ruby version targeted by JRuby 9.3 is now too old for Rubygems to run. Let's pin to the last version that works for now.

When we next have a major version bump for one of the breaking changes we're planning to make (likely around the store interface), we'll clear out support for any Ruby versions that have fallen out of security support.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,15 +12,23 @@ jobs:
 
     steps:
     - checkout
-    # JRuby <= 9.2.x is no longer supported by the `rubygems-update` gem, which now
-    # requires at least Ruby 2.6 compatibility.
+    # We need to limit the version of Rubygems that we update to for some of the older
+    # JRuby versions. Specifically:
+    #
+    #   - JRuby 9.2.x targets Ruby 2.5, and Rubygems 3.3.26 was the last version to
+    #     support that
+    #   - JRuby 9.3.x targets Ruby 2.6, and version 3.4.22 was the last version to
+    #     support that
     #
     # For now, we'll pin to the last supported version. We can drop this check once we
     # drop all those versions of JRuby.
     - run: |
-        running_old_ruby=$(ruby -e 'puts Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.6.0")')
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -12,15 +12,23 @@ jobs:
...
     # drop all those versions of JRuby.
     - run: |
-        running_old_ruby=$(ruby -e 'puts Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.6.0")')
-        if [[ "${running_old_ruby}" == "true" ]]; then 
+        running_very_old_ruby=$(ruby -e 'puts Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.6.0")')
+        running_moderately_old_ruby=$(ruby -e 'puts Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.0.0")')
+        if [[ "${running_very_old_ruby}" == "true" ]]; then
           gem update --system 3.3.26
+        elif [[ "${running_moderately_old_ruby}" == "true" ]]; then
+          gem update --system 3.4.22
```
dmagliola, line 31: This...
While I applaud the tenacity... It feels a bit sketch...

Should we let go of some old versions?

dmagliola: Ah, so your PR comment.

I think calling that a breaking change is fair enough

dmagliola (approved):

.circleci/config.yml
Sinjo: We should 100% do that when we do the next actual piece of work that causes a major version bump (we have 2-3 pending pieces of work, each of which will cause a major version bump).

I get your point about this conditional getting silly, but bumping our major version over this feels kinda pointless.

I'm also going to poke around in a separate PR and see if I can make our whole JRuby situation a bit less ridiculous. CircleCI have now entirely stopped releasing images for it (we're using their legacy images with the `circleci/` prefix and they never made one with the new `cimg/` prefix - see [this post](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034)).

PR #304, opened 2023-12-27
Add JRuby 9.4 to build matrix

Sinjo:
We've had to switch to the main JRuby Docker image as CircleCI have stopped publishing their own CI-optimised ones with extra tools installed. It turns out this isn't a problem for us as we're not dependent on much in the image.

See: https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034

This commit also switches versions 9.2 and 9.3 to the main JRuby Docker image, but not 9.1 as that breaks for some inscrutable reason. We'll drop 9.1 and all other unsupported Rubies in our next major release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,14 @@ workflows:
             - cimg/ruby:3.1
             - cimg/ruby:3.2
             - cimg/ruby:3.3
+            # We've switched to JRuby's own Docker image as CircleCI decided to
+            # stop supporting it when they deprecated their legacy images. For
+            # some reason, the official JRuby image for 9.1 fails with an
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -49,5 +49,6 @@ workflows:
             - cimg/ruby:3.1
             - cimg/ruby:3.2
             - circleci/jruby:9.1
```
electron0zero, line 51: nit: would be handy to leave a small comment on why only 9.1 is kept as is.

electron0zero (approved):

.circleci/config.yml
Sinjo: Good point! Done.

PR #308, opened 2024-05-15
Declare base64 gem dependency, ready for Ruby 3.4.

sengi:
The base64 gem is [no longer a default gem](https://docs.ruby-lang.org/en/master/NEWS_md.html#label-Stdlib+updates) in Ruby 3.4.

Add base64 to the gemspec, for forward compatibility with Ruby 3.4 and to resolve the `base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0` warning that Ruby 3.3 emits on loading client_ruby.

### DIFF
--- a/prometheus-client.gemspec
+++ b/prometheus-client.gemspec
@@ -15,6 +15,8 @@ Gem::Specification.new do |s|
   s.files             = %w(README.md LICENSE) + Dir.glob('{lib/**/*}')
   s.require_paths     = ['lib']
 
+  s.add_dependency 'base64'
+
   s.add_development_dependency 'benchmark-ips'
   s.add_development_dependency 'concurrent-ruby'
   s.add_development_dependency 'timecop'


Sinjo (approved):
Thanks for fixing this!

I'm gonna ignore that JRuby 9.1 build being sad and force the merge through. I don't know exactly what's going on there, but that version [currently gets special treatment](https://github.com/prometheus/client_ruby/blob/8b177b413f3012e430931a47152ed812c5d96798/.circleci/config.yml#L59) and I think it's time to say goodbye to it.

sengi:
Thanks @Sinjo! Was just having a look into that test failure myself. I think it may be a similar issue, this time with `tins` gem (`:test` dependency in Gemfile) using bigdecimal but not declaring it yet. Not sure why JRuby 9.1 in particular doesn't like it though.

sengi:
Sorry, scratch that last part, `tins` does declare its dependency on bigdecimal.

xeruzo:
it adds a new question for those use ruby < 2.4...
https://github.com/prometheus/client_ruby/issues/315

PR #309, opened 2024-05-17
Allow to override an existing metrics on registration

Annih:


SuperQ:
Yes, this seems like an API violation.

Annih:
Hello @Sinjo ,

Yep sorry for this PR without context, to be transparent I misclick on a GH link when I created a PR for our fork :)

Basically for internal and historical reasons we need to remove some Metrics' series from the registry, and it would even be better to atomically remove a bunch of series at once to keep a consistent set, hence my patch to replace a whole Gauge for instance.
Basically what we do with this patch is we populate a new Gauge, and replace the existing one at once, like that the next scrap will only have "consistent" metrics and labels.

I know this is not idiomatic Prometheus, but this is our usecase for now and changing all our internal clients would take way more time.

Feel free to decline this PR :)

dmagliola:
I am not 100% sure, but I believe this validation is there to prevent metrics and their labels from changing during the lifetime of yours server.

If this is useful for you, you can absolutely do this on your fork, but I think having this in the main gem can be quite dangerous for other folks using it without full context.

I would close this PR, to be honest.

However, since you are here... you mention you are using a fork of the gem. Could I ask why? It is useful for us to understand what cases our gem isn't covering so that we can improve it :-)

Annih:
> However, since you are here... you mention you are using a fork of the gem. Could I ask why? It is useful for us to understand what cases our gem isn't covering so that we can improve it :-)

I need to dig in the history, it is old (6 years old) and poorly maintained.
I guess it has been created to patch the official gem with our own tiny patches that we were sure won't be merged :D 

Actually I found a 6 years old PR from my company that aimed to remove the labels https://github.com/prometheus/client_ruby/pull/80 you can find more details about our usecase there.
The usecase is exactly the same, my new patch here is just a refactor where instead of removing obsolete labels from an existing metrics, I create a new metric and replace it in the registry. (More consistency, less diff operations)

However at that time it has been rejected too. :D I guess this will never be merged, and I don't mind. As I said, I incorrectly configured the "create PR" form.

dmagliola:
I see. Thank you very much for the context!

I'll close this PR now. Thanks!

PR #311, opened 2024-06-27
Pin `tins` dependency for JRuby 9.1

Sinjo:
The latest version of `tins` adds a dependency on `bigdecimal`, which causes JRuby to fetch and build it rather than using its built-in version. This fails on JRuby 9.1, so we need to handle that version specifically.

This gets CI back to green.

---

Once this is merged, I'm going to cut a patch release to finally get rid of the annoying warning that we emit on Ruby 3.3.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,15 +9,4 @@ group :test do
   gem 'rack-test'
   gem 'rake'
   gem 'rspec'
-  gem 'term-ansicolor'
-  # The latest version of `tins` adds a dependency on `bigdecimal`, which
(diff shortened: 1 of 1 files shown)


dmagliola:
Errrmmmm.... What is `tins`?

The gem description seems very unspecific... And I may be being thick here, but pretty sure our tests runs without it...

Can we just get rid of the gem instead of this JRuby pinning?

dmagliola:
Apparently `term-ansicolor` depends on `tins`.... I removed `term-ansicolor` too, and tests still pass locally? :shrug:

PR #312, opened 2024-06-28
Yeet `term-ansicolor` and `tins` from test dependencies

Sinjo:
I was trying to fix CI by pinning `term-ansicolor` to a specific version, but Daniel pointed out that we don't even seem to use it anywhere, so let's throw it (and its `tins` dependency) out.


dmagliola (approved):
Yeet 😂

Sinjo:
Into the sea!

PR #324, opened 2025-06-16
Do not use benchmark gem in production

Juanmcuello:
The benchmark gem is no longer a default gem in Ruby 3.5.

Add benchmark to the gemspec, for forward compatibility with Ruby 3.5 and to resolve the `benchmark was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.' warning that Ruby 3.4 emits on loading client_ruby.

https://docs.ruby-lang.org/en/master/NEWS_md.html#label-Stdlib+updates

### DIFF
--- a/lib/prometheus/middleware/collector.rb
+++ b/lib/prometheus/middleware/collector.rb
@@ -1,6 +1,5 @@
 # encoding: UTF-8
 
-require 'benchmark'
 require 'prometheus/client'
 
 module Prometheus
@@ -34,6 +33,12 @@ module Prometheus
 
       protected
 
+      def realtime
+        start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        yield
+        Process.clock_gettime(Process::CLOCK_MONOTONIC) - start_time
+      end
+
       def init_request_metrics
         @requests = @registry.counter(
           :"#{@metrics_prefix}_requests_total",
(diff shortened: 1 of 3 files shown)


dmagliola:
When I saw this I thought "surely this should be a development dependency, for our benchmarks, which almost no one will encounter", but we're actually using Benchmark in our Collector Middleware to measure response time...

I do wonder whether we should switch to the much more common `Process.clock_gettime(Process::CLOCK_MONOTONIC)` instead and lose this depenency.

@Sinjo @Juanmcuello any thoughts?

Juanmcuello:
Yeah, I saw that too. I thing dropping Benchmark would be better. There are some specs that depend on the lib, but I think everything could be replaced with something like this:

```ruby
# lib/prometheus/middleware/collector.rb   
def realtime
  start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  yield
  Process.clock_gettime(Process::CLOCK_MONOTONIC) - start_time
end

# spec/prometheus/middleware/collector_spec.rb
expect(app).to receive(:realtime).and_yield.and_return(0.2)
```

I can try a PR if you think that's the way to go.

Juanmcuello:
But I see that even with these changes, we would still depend on Benchmark for testing, so we would have to keep it as a dev dependency anyway.

Sinjo (approved):

Sinjo:
Thank you @Juanmcuello! I'll get a release out this weekend.

PR #330, opened 2026-02-05
Fix Ruby 4.0 compatibility: replace CGI with URI in DirectFileStore

dannote:
CGI.parse and CGI.escape methods were removed from the `cgi` gem in Ruby 4.0. The `cgi` gem now only provides `CGI.escape`/`CGI.unescape` for HTML escaping, and `CGI.escapeURIComponent`/`CGI.unescapeURIComponent` for URI components.

This PR replaces:
- `CGI::parse(query_string)` → `URI.decode_www_form(query_string).to_h`
- `CGI::escape` in query string building → `URI.encode_www_form`

The `uri` module is part of Ruby's standard library and doesn't require an additional dependency.

**Testing:**
- All existing specs pass
- Tested manually with Ruby 4.0.1 in production environment

### DIFF
--- a/lib/prometheus/client/data_stores/direct_file_store.rb
+++ b/lib/prometheus/client/data_stores/direct_file_store.rb
@@ -1,5 +1,5 @@
 require 'fileutils'
-require "cgi"
+require "uri"
 
 module Prometheus
   module Client
@@ -133,12 +133,9 @@ module Prometheus
               begin
                 store = FileMappedDict.new(file_path, true)
                 store.all_values.each do |(labelset_qs, v, ts)|
-                  # Labels come as a query string, and CGI::parse returns arrays for each key
-                  # "foo=bar&x=y" => { "foo" => ["bar"], "x" => ["y"] }
-                  # Turn the keys back into symbols, and remove the arrays
-                  label_set = CGI::parse(labelset_qs).map do |k, vs|
-                    [k.to_sym, vs.first]
-                  end.to_h
+                  # Labels come as a query string
+                  # "foo=bar&x=y" => { foo: "bar", x: "y" }
+                  label_set = URI.decode_www_form(labelset_qs).to_h.transform_keys(&:to_sym)
 
                   stores_data[label_set] << [v, ts]
                 end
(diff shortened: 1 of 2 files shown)


Sinjo:
Sorry this has taken me so long to get to.

I definitely like the idea of using `URI` since it's still part of the Ruby stdlib. It lets us avoid the extra dependency I introduced in #327.

I wanted to be really sure that we weren't going to run into any unintended side-effects with weird Unicode characters. It struck me that computers are fast and there aren't that many characters, so it's easy to exhaustively test the two serialisation/deserialisation approaches:

```ruby
#!/usr/bin/env ruby

require 'cgi'
require 'uri'

# Encoding differences
(0x0000..0x10FFFF).map { |i|
  begin
    s = i.chr(Encoding::UTF_8)
  rescue RangeError => _
    next
  end

  [
    i,
    s,
    CGI::escape(s),
    URI::encode_www_form([s]),
  ]
}.compact.filter { |a|
  a[2] != a[3]
}.each { |a|
  puts "Encoding differed for codepoint: #{a[0]} CGI: #{a[2]} URI: #{a[3]}"
}

# Round-trip failures
(0x0000..0x10FFFF).map { |i|
  begin
    s = i.chr(Encoding::UTF_8)
  rescue RangeError => _
    next
  end

  [
    i,
    s,
    CGI::parse(CGI::escape(s)).first.first,
    URI::decode_www_form(URI::encode_www_form([s])).first.first,
  ]
}.compact.filter { |a|
  a[1] != a[2] || a[1] != a[3]
}.each { |a|
  puts "Round-trip differed for codepoint: #{a[0]} CGI: #{a[2]} URI: #{a[3]}"
}
```

which results in:

```
$ bundle exec ./my_utf8_test.rb
Encoding differed for codepoint: 42 CGI: %2A URI: *
Encoding differed for codepoint: 126 CGI: ~ URI: %7E
```

Most importantly, round-tripping works equally well in both approaches. I don't think the two characters they encode differently are problematic for our usage.

I'm going to merge this and get a release out.

Sinjo (approved):

PR #332, opened 2026-04-15
Add quantiles to summaries

cyberdelia:
Calculate quantiles in summaries, this makes `Summary` actually useful for this client. This reuse the default parameters from the Go client to be somewhat consistent.

### DIFF
--- a/lib/prometheus/client/formats/text.rb
+++ b/lib/prometheus/client/formats/text.rb
@@ -50,6 +50,11 @@ module Prometheus
           end
 
           def summary(name, set, value)
+            value.each do |k, v|
+              next if k == "sum" || k == "count"
+              yield metric(name, labels(set.merge(quantile: k)), v)
+            end
+
             l = labels(set)
             yield metric("#{name}_sum", l, value["sum"])
             yield metric("#{name}_count", l, value["count"])
--- a/lib/prometheus/client/registry.rb
+++ b/lib/prometheus/client/registry.rb
@@ -45,11 +45,17 @@ module Prometheus
                              store_settings: store_settings))
       end
 
-      def summary(name, docstring:, labels: [], preset_labels: {}, store_settings: {})
+      def summary(name, docstring:, labels: [], preset_labels: {},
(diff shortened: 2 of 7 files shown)


dmagliola:
How does this work with aggregation of Data Stores from different processes?

The library used to have Quantile support, and we removed it when we added support for multi-process Web servers.

My understanding at the time is that quantile are fundamentally non-aggregatable (at least not without storing every observation which is not feasible). It's possible this is because of the `quantile` gem design, which you're not using, and not a fundamental limitation, but i'd like to understand how this works when the observations from different processes get aggregated.

[Explanation on the issue with aggregating quantiles in our proposal](https://github.com/prometheus/client_ruby/issues/94#:~:text=2.%20Remove%20Quantile%20calculation%20from%20Summary)

[Removal of quantile support, also with explanation](https://github.com/prometheus/client_ruby/pull/95/changes/ce6d44dea95f8af7f4d772c68cd2775f6668ec2d).

cyberdelia:
I failed to do a bit of homework. This doesn’t behave much better than the previous implementation in that matter.

But for use cases where the trade-off of not using a multi-process web server is acceptable, collecting quantiles in summaries is very useful (rather than having them do almost nothing).

I’m happy to add a guard preventing declaring objectives in summaries for multi-process compatible data stores alongside some warning in the documentation. I’d be happy to do that with the previous quantile implementation as well.

SuperQ:
It's easier, and more accurate, to use a Histogram.

cyberdelia:
@SuperQ It’s true for most cases, but for rarely emitted metrics, the histogram bucketing gets in the way in a big way.

SuperQ:
Bucket storage is no longer an issue with native histogram storage.

cyberdelia:
@SuperQ But they are also not supported by this client.

SuperQ:
It's not required that the client supports it. See the [`convert_classic_histograms_to_nhcb`](https://prometheus.io/docs/prometheus/latest/configuration/configuration) feature.

